PR #2, opened 2024-04-07
fixed character flaws

TiberianEuan:
only problem with it is the fact that you can just choose random or no flaw, but for right now this is a mere bandaid fix
image

BUT
FINALLY
I CAN STOP BEING A FUCKING CYCLOPS


TiberianEuan:
holey moley

PR #20, opened 2024-04-09
gives skeletons their language priviledges back

TiberianEuan:
gives skeletons their language priviledges back (and makes them able to reattach limbs like in Goon)


TiberianEuan:
im merging this because other admins agree

PR #23, opened 2024-04-10
Makes all classes always show up when chossing class

K4rlox:
no longer luck based


bobjension:
this is probably bad cuh uh there's no reason for people not to pick the better classes

Xkallubar:
Huzzah

YosemiteYam:
Maybe limit the rarer class slots like the upgraded versions of hunter (ranger), monk, crusader etc

Henri9576:
Having to either kill myself and go through underworld or wait until the next round to do a gimmick that I have in mind for a specific class due to pure RnG is not a good gameplay feature.

Good change.

K4rlox:
> Yeah this generally just seems bad... might as well just remove the worse classes.

im gonna buff and nerf some classes after this gets merged

ChungusGamer666:
Awful for balance but I don't feel comfortable nuking this PR into the stratosphere

ChungusGamer666:
didnt mean to close it OOPS yeah no

Lauwma:
I like this PR but not for the same reasons. What I liked with this change is that it allowed character consistency. It's weird to come up one day as a ranger and another as a mage. Character consistency is good because it allows for organic evolutions. Maybe your character is an adventurer now but after an event they're gonna become a Towner or a town guard. 

Another alternative that fixes this problem would be to make a character selection happen after you chose your subclass.

ChungusGamer666:
Closing this PR because the dev consensus seems to be: Not good, impacts balance and incentivizes bad player behavior (conformism).
I would consider expanding class selection by some margin but this is too much.

Will reopen later if a bigger belly button (emoney or jamie) tells me to reopen it.

PR #41, opened 2024-04-12
Major Map Changes, Fixes Patrons Qdel'ing, Removes ESL PQ Penalties, More

Cdey78:
### **MAP CHANGES:**
) Changes the STEWARD to take the place of the manor gatehouse. 
-- IMPORTS are moved to the far south-east of town. If you follow the street straight south of the manor house and take a RIGHT at the T section, you can follow the path down to the import docks. It's best you send the guards, lest the homeless take advantage of the lord's new stock!

) Fixes beggars and orphans from spawning in the sewers with no light.

) The smithy bridge is reintroduced, no longer is the divide between our hearts.

) A millstone has been added to the farm outside of town.

### **OTHER**
) PQ no longer lowers on a misspelling.
) Clerics and Monks would Qdel() the player's patron when it did not fall under the allowed patrons. They now properly re-set their patron.
) Setting your character flaw in the lobby now sends the description in the chatbox, again.
) Deletes the unused SEX subsystem.

This one compiles and works locally. Should be ready for merge.


Cdey78:
I forgot to fix the commits, I'm a mong

YosemiteYam:
Does this fix the gatehouse acid pit lever not working

PR #46, opened 2024-04-13
Char Pref Screen Break Fix Attempt One + one error cleanup

JTGSZ:
attempt number 1 to do something about the character pref screen break because its inconsistent to get it out


JTGSZ:
didn't work

PR #63, opened 2024-04-14
Fix pickpocketing

th6795:
Yeah so there's an existing bug with pickpocketing allowing users to "succeed" at pickpocketing without actually stealing anything.

This introduces the following changes:
- Users now receive a descriptive message when they try to pickpocket a damage zone that is "unpickpocketable"
- For damage zones that have multiple slots that they steal from (such as torso stealing from back left and back right), only slots that have items in them will are considered when picking what item is stolen
- When a user attempts to pickpocket a "pickpocketable" damage zone whose assigned theft slots are all empty, the user will receive a descriptive measure indicating that they didn't find anything.
- Users were able to pickpocket while having an item in their active hand. For one, this doesn't make any sense. For two, it caused the stolen item to just drop on the ground- super awkward for thief.


code/_onclick/other_mobs.dm
```diff
@@ -378,24 +378,36 @@
 					var/list/stealablezones = list("chest", "neck", "groin", "r_hand", "l_hand")
 					var/list/stealpos = list()
 					if(stealroll > targetperception)
+						if(U.get_active_held_item())
+							to_chat(src, "<span class='warning'>I can't pickpocket while my hand is full!</span>")
+							return
 						if(!(zone_selected in stealablezones))
-							to_chat()
+							to_chat(src, "<span class='warning'>What am I going to steal from there?</span>")
```
ChungusGamer666, line 385: lol

PR #66, opened 2024-04-14
Fixes for missing skeleton sprites, fixes for ambush mobs not getting clothing, uncomments zombie stat change, makes zombie more streamlined

JTGSZ:
yeeeep


code/modules/mob/living/carbon/human/npc/bum.dm
```diff
@@ -9,12 +9,12 @@ GLOBAL_LIST_INIT(bum_aggro, world.file2list("strings/rt/bumaggrolines.txt"))
 	dodgetime = 30
 	flee_in_pain = TRUE
 	possible_rmb_intents = list()
-	static_npc = TRUE
+
 	wander = FALSE
 
 /mob/living/carbon/human/species/human/northern/bum/ambush
 	aggressive=1
```
Osmos123, line 16: Hello nice changes, does this part need to be changed to "agressive = 1" like the others to work?

JTGSZ: Up to you. If you hit a bum that isn't part of an ambush they get pissed, in the ambush they are already pissed.

PR #70, opened 2024-04-14
Paladin Nerf

Osmos123:
Reduces the sword skill of paladins from expert (lvl 4) down to skilled (lvl 3) to be more in line with other adventurer classes starting skill level. Still an overtuned class that starts with the best helmet variant and healing. Hopefully this change will at the very least delay the murderboning of the entire nobility and beggars. 

The intelligence is reduced to try to slow down training speed (although I rarely see paladins train they just go ham at the start). The paladins still have a above average physical attribute bonus and considering the current playstyle of paladins are well-armed marauders (basically crusaders).

### DIFF
--- a/code/modules/jobs/job_types/roguetown/adventurer/types/combat/donator/templar.dm
+++ b/code/modules/jobs/job_types/roguetown/adventurer/types/combat/donator/templar.dm
@@ -26,7 +26,7 @@
 		H.mind.adjust_skillrank(/datum/skill/combat/bows, 2, TRUE)
 		H.mind.adjust_skillrank(/datum/skill/combat/wrestling, 3, TRUE)
 		H.mind.adjust_skillrank(/datum/skill/combat/unarmed, 3, TRUE)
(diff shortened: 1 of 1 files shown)


Osmos123:
Yep that would be a better option. Nerf was more intended if the current paladin infinite slots was kept.

Pockets-byte:
this makes more sense. paladin's were a PTW class, and just serve to replace cleric as a better variant. 
it seems like they could use some distinction or as suggested, be shifted to soldiers in the service of the church to separate them from being a species and gender locked upgrade to clerics

Hmeister-real:
> shifted to soldiers in the service of the church 

churches in real medieval times were like, really powerful, would be fun for the church to have power similar to the king. It would give Priests more reason to exist other than being better acolytes and it would make it more common for sinners to be punished by the church rather than by the guards which really really need to happen (playing as a guardsmen is a stressful confusing nightmare imo)((also maybe there could be a dungeoneer for the church? like beating the sin out of the sinful, pure ideasguy because on the earlier map the church had a cell with confessions and a flail))

PR #74, opened 2024-04-14
[QOL - READ DESC] [CLOSED BECAUSE SPLIT INTO NEW PRs] Lets pilgrims pick all worker classes, flint (&steel) for adventurers/pilgrims, tweaks equipment to make more sense.

ariaworld:
Gives adventurers and pilgrims satchels, making each class closer to being balanced with one another, also gives them flint & steel for torches. 
Pilgrims also spawn with more money than adventurers to counter-act the fact adventurers have stronger options.
Pilgrims can now choose all kinds of pilgrim worker classes, without the 5-slot limit.


JohnChiffir:
I still don't understand how changing every backpack to a satchel will make a balance better, especially that most of roles that have a backpack are pilgrims... and why did ranger lose his lamp though

lovegreenstuff:
I'm gonna be honest, I feel like this should have been 4 separate PRs.

code/modules/jobs/job_types/roguetown/adventurer/types/combat/donator/grenzelhoft.dm
```diff
@@ -13,7 +13,9 @@
 	wrists = /obj/item/clothing/wrists/roguetown/bracers
 	belt = /obj/item/storage/belt/rogue/leather
 	armor = /obj/item/clothing/suit/roguetown/armor/gambeson
-	beltr = /obj/item/storage/belt/rogue/pouch/coins/poor
+	beltr = /obj/item/storage/belt/rogue/pouch/coins/mid
```
LudwigVonChesterfield, line 16: isn't the point of mercenaries that you hire them for money

ChungusGamer666:
interesting PR, i'll have to look into this later and see if it aligns with my vision for pilgrims

Henri9576:
If I may suggest something for this, maybe make it so you can pick all pilgrim classes instead of the random five, with a small chance for one of them to be legendary?
It make sense for adventurer to not have this, since some classes are clearly better, but for pilgrim, people can just choose what they want to specialize in instead of leaving it in the hand of rng.

PR #76, opened 2024-04-14
Lesser Heal Obliteration

Osmos123:
Its a crutch that circumnavigates sleeping in a bed. Removed most of the healing damage aspect, can only be used to stabilize wounds such as bleeding. Bed healing gameplay is essential as it is a moment of vulnerability one has to experience (more rp with group less murder hobo). This should also tone down the strength of paladin/cleric while keeping the church viable.

### DIFF
--- a/code/modules/spells/roguetown/cleric.dm
+++ b/code/modules/spells/roguetown/cleric.dm
@@ -85,12 +85,12 @@
 			var/mob/living/carbon/C = target
 			var/obj/item/bodypart/affecting = C.get_bodypart(check_zone(user.zone_selected))
 			if(affecting)
-				if(affecting.heal_damage(20, 20, 0, null, FALSE))
+				if(affecting.heal_damage(0, 0, 0, null, FALSE))
 					C.update_damage_overlays()
 				if(affecting.heal_wounds(50))
 					C.update_damage_overlays()
 		else
-			target.adjustBruteLoss(-5)
+			target.adjustBruteLoss(-1)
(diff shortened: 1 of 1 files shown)


SELFHELL:
-1 From me personally, I get the intent but the implications of completely removing the heal from the spell are icky. I'd just personally remove it from the roundstart kit of paladins if you find it that problematic. Also just reducing the healing amount to zero is a lazy way of removing it.

BlueMemesauce:
Paladin was op on purpose, it was a donator role

code/modules/spells/roguetown/cleric.dm
```diff
@@ -85,12 +85,12 @@
 			var/mob/living/carbon/C = target
 			var/obj/item/bodypart/affecting = C.get_bodypart(check_zone(user.zone_selected))
 			if(affecting)
-				if(affecting.heal_damage(20, 20, 0, null, FALSE))
+				if(affecting.heal_damage(0, 0, 0, null, FALSE))
```
LudwigVonChesterfield, line 88: what's the point of leaving this function call here if it does nothing

ChungusGamer666:
i kinda agree with this PR but it seems like most people dont

the medical system right now is literally broken and things like surgery are completely nonfuctional, and bed healing doesnt cure stuff like broken bones etc so yeah this pr kinda stings

Cdey78:
We can come back to this when medical is in a better state.

ChungusGamer666:
what cdey said... ripbozo

PR #90, opened 2024-04-15
quick fix for cloth crafting

cornercube:
gambeson too strong to allow it to be crafted, oops
have some robes instead

### DIFF
--- a/code/modules/roguetown/roguecrafting/sewing.dm
+++ b/code/modules/roguetown/roguecrafting/sewing.dm
@@ -78,12 +78,12 @@
 
 /* craftdif of 2+ */
 
-/datum/crafting_recipe/roguetown/sewing/clotharmor
-	name = "gambeson"
-	result = list(/obj/item/clothing/suit/roguetown/armor/gambeson)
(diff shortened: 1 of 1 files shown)


Pockets-byte:
part of the core of rougetown was the ability to build up from nothing. what about this one item makes it so special that it has to be close to adminspawn only- to my knowledge, gambesons are light armor with close to if not the exact same stats as leather armor, I can't find it's stats in the code right now, but removing the item is not the proper approach to fixing an issue

cornercube:
> part of the core of rougetown was the ability to build up from nothing. what about this one item makes it so special that it has to be close to adminspawn only- to my knowledge, gambesons are light armor with close to if not the exact same stats as leather armor, I can't find it's stats in the code right now, but removing the item is not the proper approach to fixing an issue

removed by request on discord, on the grounds it was the strongest light armor in the game and had to be purchased from the merchant for a reason
it was 40 armor, covered torso, arms, legs, vitals
leather armor was 25 armor, covering only torso, vitals

but theres a PR up right now that will re-add them, so its all moot

PR #95, opened 2024-04-15
Knocking is back on the menu boys!

panvxv:
Enables knocking people out with blunt weapons and tools.
Really experimental as no idea was this removed.
(Was ON during RogueTown era)

### DIFF
--- a/code/modules/surgery/bodyparts/wounds.dm
+++ b/code/modules/surgery/bodyparts/wounds.dm
@@ -343,9 +343,6 @@
 				owner.next_attack_msg += " <span class='crit'><b>Critical hit!</b> [owner] is knocked out!</span>"
 				owner.flash_fullscreen("whiteflash3")
 				owner.Unconscious(600)
-				if(owner.client)
-					winset(owner.client, "outputwindow.output", "max-lines=1")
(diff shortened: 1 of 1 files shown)


code/modules/surgery/bodyparts/wounds.dm
```diff
@@ -261,7 +261,7 @@
 					if(owner.client)
```
LudwigVonChesterfield, line 261: this won't work with the current "T" hotkey :eyes:

panvxv: There is no T hotkey being used here. What do you mean?

LudwigVonChesterfield: I believe the intention of three lines of code (261-263) is to output/prevent user from typing by changing the max-lines of the command line on the bottom right

but because people use the "say" hotkey to communicate it won't work.

panvxv: I am testing it and I can't talk when I am uncounscious

LudwigVonChesterfield: 👍

ChungusGamer666:
If this goes back we can revert it idc

PR #99, opened 2024-04-15
[ADMIN-APPROVED] Templar job role for Church (original Paladin adventurer role), nerfed Paladin for adventurers - read desc

ariaworld:
Paladin adventurers start without the templar helmet and churn spell, and have lesser heal + burial rites. -1 sword skill and -1 crossbow skill, no more praying for new spells.

Templar role is the same as old Paladin except it has the extra burial rite spell, and no more praying to get more spells.
(max 2 slots)

(Effectively solving the problem of paladins being unbalanced, so we can aim for having most adventurers classes able to be played and not a random roll, except for special classes, but that's another story)

1st and 2nd pic: Templar, showing no praying ability, but 3 spells.

3, 4, and 5 showing Paladin, no templar helmet, only 2 spells, no pray ability.

templar1
templar2
paladin
paladin2
paladin3


Osmos123:
I wonder if you could adjust the charge max of the miracle to 20 seconds to account for how powerful it is compared to the lesser miracle that 10 seconds. At least until the devotion bug gets fixed.

Xkallubar:
Their secondary spell should be dependant on their patron, as with clerics. Burial Rites is an exclusively Necra spell.

YosemiteYam:
The templar description is the same as the paladin description

"Templars are former noblemen and clerics who have dedicated themselves to great combat prowess. Often, they were promised redemption for past sins if they crusaded in the name of the gods."

"Paladins are former noblemen and clerics who have dedicated themselves to great combat prowess. Often, they were promised redemption for past sins if they crusaded in the name of the gods."

It should be different and less-militaristic, it implies like they're about to embark on a holy crusade in any moment. We don't want the church to start acting like some dumbfuck military faction and start a holy revolt every round, it should emphasize that they're there to protect the church and fight against heresy or unholy creechers lurking in town at most.

YosemiteYam:
"Warriors who have forsaken wealth and title in lieu of service to the church, due to either zealotry or a past shame. They guard the church and its priest, while keeping a watchful eye against heresy and nite-creechers. Within troubled dreams, they wonder if the blood they shed makes them holy or stained"

Keeps it inline with the other role descriptions that are abit ironic/grimdark

Osmos123 (commented):
I would say keep this line of code for the miracle for balance purposes, maybe even add it in to the lesser miracle. (I believe it prevents self healing, at the very least when playing as paladin my stronger miracle spell had no effect on me.)

PR #101, opened 2024-04-15
Decorated Swords

Diegoflores31:
-removed unnecesary skill checks on every single blackmith recipy.
-you can now "engrave" existing weapons to turn them into their decorated variants.
-organized the crafting .dm a little


JTGSZ:
I reviewed it, zweihander mats is weird but its the same as it was before you put this together so its fine.

PR #107, opened 2024-04-16
Replace ass slapping with belly popping

NGGJamie:
Slapping asses is still in the code and is a fast trip to banville. Instead, players can now pop bellies.


DorfGuy1:
Bad

PR #108, opened 2024-04-16
Replaces all instances of var/ in proc arguments

Xander3359:
Uses the following regex:
``(^/[\w/][^ \t]+\(.*)(var/|, ?var/.*)(.*)``
to search for all instances of .../proc/proc_name(var/proc_arg)
And performs the following substitution:
$1$3

Because var/ is unecessary and removing these manually is unreasonable.

### DIFF
--- a/code/__HELPERS/heap.dm
+++ b/code/__HELPERS/heap.dm
@@ -33,7 +33,7 @@
 		Sink(1)
 
 //Get a node up to its right position in the heap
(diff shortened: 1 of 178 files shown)


ChungusGamer666:
wholesome pr

JTGSZ:
I skimmed it, var/ in proc params can cause things to disjoint sometimes in the first place when someone forgets something, so I salute you for the janny work.

PR #110, opened 2024-04-16
Removes Special Messages Relating to Slaps

NGGJamie:
Title.


JTGSZ:
funny things will return into that proc one day

PR #111, opened 2024-04-16
Buffs needle for wound tending (see body)

Xander3359:
- Condenses a few checks
- a needle is now used only when you finish tending a wound
- you now get medicine experience when you close a wound
- Nerfs the amount of progress you get from medical skill (now it heals +5 per level isntead of +30, keeping in mind it requires 100 to heal)


ChungusGamer666:
Godly PR will take a look later

emoats18:
Will merge will need to be tested by our gracious players

PR #134, opened 2024-04-16
Sunders unused lavaland ruin files

AMyriad:
Deleted all the lavaland ruin .dmm files and others supporting said ruins, lemme know if you want me to hold fire


Xander3359:
So wait are we touching TG code or what's going on? We should decide on wether or not we want to keep TG code + stay modular, or axe TG code completely at this point.

AMyriad:
> So wait are we touching TG code or what's going on? We should decide on wether or not we want to keep TG code + stay modular, or axe TG code completely at this point.

These are good concerns. I haven't been informed of what the official maintainers want to do with the mountains of unused code, but this is my take:
Blackstone/Roguetown seems to use next to no original SS13 content. All I see used are its systems, helpers, and defines, which make sense. But stuff like vending machines, laser guns, robots, cyborgs, AIs, space ruins, lavaland, stations, etc etc, are just dead weight. They take up file space and if anything, hurt performance (which is currently a big issue).

I'm trying to be as surgical as possible to not meddle with what systems Blackstone relies on, so I'm gonna be atomizing these PRs  for the sake of trying not to break stuff + making it easy to review.

TL;DR unused code is bad code. If we want to reuse this unused code for cool stuff later on, that's great, but I don't believe it'd be hard to look at other repositories for code.

Arturlang:
What we should do is un-tick most TG files, this would also lead for far faster compile times too

Xander3359:
> What we should do is un-tick most TG files, this would also lead for far faster compile times too, doesn't stop us from being modular, that

If we care about modularity, we should rebase to TG current and just rebuild roguetown on top of it.
It would be a better idea than slowly updating invidual TG files as we need to.

JTGSZ:
developed conflicts

AMyriad:
> developed conflicts

conflicted with the space ruin thing, should be a-okay now

PR #29, opened 2024-04-11
Dark Vision and You: How I learned to hate Elves even more

SELFHELL:
Elves and their halfblooded kin have regained the blessing of Darksight, no longer will mild inconveniences like losing a torch bother the superior race(s)

Dwarves now heal rather quickly from alcohol consumption

PR #135, opened 2024-04-16
Night vision fix

### LINKED ISSUE #131: Drows dont get dark vision as intented from #29
Apparently they should have it as a trait but unfortunately they do not.It dosent appear even when rightclicking skills (although im not really sure it should even appear there)

steps to reproduce:
go in dark
DARKNESS

image

I would like to point out when I was zombiefied it appeared I did have darkvision of some degree since i was able to see in the dark.

BlueMemesauce:
fixes #131

#29 had it the opposite way around

also reduced range otherwise it would be too op

half elf
image

elf
image


SELFHELL:
oh, beat me to the fix; good shit.

PR #139, opened 2024-04-16
Bombs! Explosive crafting and small buff to Dwarf Bomber

K4rlox:
Dwarf bomber loadout exists, it has only 1 bomb and a flint, and gets really high cratfing skill

issue being? you cant craft bombs

now you can craft a bomb using
1 cup or bottle
1 coal
1 cloth

also the dwarf bomber gets 1 more bomb, for being a bomber

### DIFF
--- a/code/modules/jobs/job_types/roguetown/adventurer/types/combat/dbomb.dm
+++ b/code/modules/jobs/job_types/roguetown/adventurer/types/combat/dbomb.dm
@@ -16,7 +16,7 @@
 	armor = /obj/item/clothing/suit/roguetown/armor/chainmail/iron
 	wrists = /obj/item/clothing/wrists/roguetown/bracers/leather
 	backl = /obj/item/storage/backpack/rogue/backpack
-	backpack_contents = list(/obj/item/bomb = 1, /obj/item/flint = 1)
+	backpack_contents = list(/obj/item/bomb = 3, /obj/item/flint = 1)
 	if(prob(50))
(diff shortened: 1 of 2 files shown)


LudwigVonChesterfield (approved):

YosemiteYam:
Thats way too cheap to make bombs, coal is in abundance just from people mining for iron, you can get/make cloth and cups for free. There needs to be an actually rare resource in the ingredients that stops grenade spam being a reality, especially from what I noticed grenades are NOT weak at all, and frequently burn down and entire group or building when a bored dwarf throws it in a crowd. Otherwise you can just make an endless supply of grenades, and with the fact that you cant put out fire on other people this will just grief the server

Arturlang:
Being made out of coal is fine, but it depends on how effective those bomb are, maybe make it need metal than a mug?

JTGSZ:
There is already problems with people attempting mass-killing on spawn right now. I support more bombs on the dwarf, and I support the dwarf creating more bombs, but giving them a easy option for mass aoe isn't wise.

K4rlox:
> There is already problems with people attempting mass-killing on spawn right now. I support more bombs on the dwarf, and I support the dwarf creating more bombs, but giving other guys outside of said dwarf a easy option for mass aoe isn't wise.

you need crafting skill

CVSthePharmacy:
why not make an entire new skill called bombcrafting for the recipe, should make it more balanced as nobody else will be able to even make it

Issue #140, opened 2024-04-16
Templars don't spawn with church keys and can't pray for devotion


BadAtThisGame302:
Title

PR #146, opened 2024-04-16
The solution to Paladins and Templars, part 3.

ariaworld:
Adds max_devotion to Templars and Paladins so they can pray again to cast spells, balances fighting classes such as warrior and paladin to be closer to eachother.


Osmos123:
image

Its so over paladin bros..

Osmos123:
Also could you add for paladin and templar to prevent them from being recruitable in a peasant rebellion? It does not make sense when both derive from nobility and both nobilty and church are 100% reliant on each other to maintain control of the lower-class population.

https://github.com/Blackstone-SS13/BLACKSTONE/blob/9c912356016bbc62eabfe98f07a212ddd69f98fe/code/modules/antagonists/roguetown/villain/peasantrebel.dm#L4

Henri215:
Link the issue https://github.com/Blackstone-SS13/BLACKSTONE/issues/140 so it can be closed

PR #148, opened 2024-04-16
Church lighting fix + Farm door locks + Church pew dir fix

TopHatPenguin:
- adds two sconces to the main altar for improved lighting
- slightly moves the sconces west and east in the church to provide more lighting cover

Change: https://i.gyazo.com/cc965ccc45e300dfe38b40174aeddee5.png

- Farm doors now actually have locks so their keys will work
- Fixes church pews to all share the same DIR


silverplatedelta (approved):
Code looks fine. Map properly loads and the church lighting is noticeably improved.

PR #149, opened 2024-04-17
Unlimits towner slots

zakanater19:
allows players to latejoin in the town reliably and not lost in the woods


Henri215 (approved):

ChungusGamer666:
I don't want this change neither does emoney
We only unlimited beggar slots because it's a funny shitty job

zakanater19:
Fair enough

PR #157, opened 2024-04-17
More flexible floor crafting

Osmos123:
Allows building over existing floors (before was only allowed for dirt). Been dwarf fortressing and it hurts my soul that we cannot build over the existing floors only the dirt.

image


LudwigVonChesterfield:
this will make for a great meme of 5 layers of stone road. I think this is cool

PR #159, opened 2024-04-17
Literally 1984: Stops spirits from fighting each other

Contrabang:
Spirits are now made of ghosty stuff... or whatever. They can't touch each other now. Requested by Maintainers & Host
image


Arturlang:
using a pacifism trait with a desc like that is not good.

PR #170, opened 2024-04-17
Wealth horde working for all items

panvxv:
Made so items inside of treasury room will generate small amount of mammons over time - passive income.

I wanted it to only work on items in bounties file but I am not a coder :)
Chose nuke option and it will generate mammons from every item.


JTGSZ:
Closing because trying to reduce the amount of mammons in the economy right now

NGGJamie:
With where the economy is now, I would like to run this and see how it goes. (when we're out of hardware team  hell)

PR #173, opened 2024-04-17
Gives gravediggers the poor coin pouch instead of empty pouch

Contrabang:
Gives gravediggers the poor coin pouch instead of empty pouch. It helps them bury the dead.


Henri215 (approved):
Lewcs good to me :)

DGamerL (approved):
Do the selfmerge 🔫

PR #177, opened 2024-04-18
Fix shrunken lamps

Contrabang:
Fix shrunken lamps. Makes them look as close as possible to the old lamps


BlueMemesauce:
it's not just shrunken lamps, its all lamps/torches have been decreased by about half

PR #179, opened 2024-04-18
Why do we need lavaland ruins

RealityOverseer:
There is no standard format for the PR's and that pains me a little.

Anyways, this removes the entire lavaland ruin folder because... why do we need it?

I tested it via compiling with no errors and made sure all references to the files deleted was purged.


NGGJamie:
Possibly a copy of #134 though I do see differences in file count.

RealityOverseer:
After actually looking at the PR's instead of just doing, it does appear that this is a duplicate. Difference is that #134 is also modifying hierophant. Thus this can probably be closed.

PR #181, opened 2024-04-18
Lighting added to: garrison and tavern + merchant doors locked

TopHatPenguin:
Title.

- Additionally the Navi room's window has been replaced with a wall as it was too easy to break into. 
- fixes the invis table in the inn
- Correctly labels the three town houses and the corresponding keys (w/locks)


JTGSZ:
Merging before conflicts develop

PR #188, opened 2024-04-18
Toll Fix

NGGJamie:
Tolls do not seem to be spawning, give the spirits a break until it's fixed.

Only expecting to run this as a TM until a real fix is done.


NGGJamie:
The list used to count coins was being filled with nulls which threw off the count. Because this list is never iterated over, I converted it to a number instead. It now works as intended.

PR #193, opened 2024-04-18
Fix the required job system

### LINKED ISSUE #161: Low PQ king
Having low pq and setting king to low is meant to allow the player to join as king, if there aren't any other rollers. Currently the game won't pick you at all even with the role set to low. This is quite easy to test in private hosting as you just join set king low and force start and watch as you aren't king. Related to #104

BlueMemesauce:
fix #161

### DIFF
--- a/code/controllers/subsystem/job.dm
+++ b/code/controllers/subsystem/job.dm
@@ -560,8 +560,8 @@ SUBSYSTEM_DEF(job)
 				if(!job.special_job_check(player))
 					continue
 
-				// If the job isn't filled.
-				if((job.current_positions < job.spawn_positions) || job.spawn_positions == -1)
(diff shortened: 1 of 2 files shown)


code/controllers/subsystem/job.dm
```diff
@@ -212,9 +212,6 @@ SUBSYSTEM_DEF(job)
 			JobDebug("GRJ incompatible with PATREON LEVEL, Player: [player], Job: [job.title], Race: [player.client.prefs.pref_species.name]")
 			continue
 
-		if(get_playerquality(player.ckey) < job.min_pq)
```
BlueMemesauce, line 215: this was here twice btw thats why i removed it

PR #195, opened 2024-04-18
Prevent Spirits From Stripping Each Other

NGGJamie:
Currently, spirits can steal tolls from each other via the strip menu. This prevents that.

### DIFF
--- a/code/modules/mob/living/living.dm
+++ b/code/modules/mob/living/living.dm
@@ -1184,6 +1184,10 @@
 	if(check_arm_grabbed(active_hand_index))
 		to_chat(src, "<span class='warning'>Someone is grabbing my arm!</span>")
 		return
+	
+	if(istype(src, /mob/living/carbon/spirit))
+		to_chat(src, "<span class='warning'>A field of magic surrounds \the [what], preventing you from touching it.</span>")
+		return
 
 	who.visible_message("<span class='warning'>[src] tries to remove [who]'s [what.name].</span>", \
 					"<span class='danger'>[src] tries to remove my [what.name].</span>", null, null, src)
--- a/code/modules/mob/living/living.dm
+++ b/code/modules/mob/living/living.dm
@@ -1186,7 +1186,7 @@
 		return
 	
 	if(istype(src, /mob/living/carbon/spirit))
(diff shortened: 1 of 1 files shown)


code/modules/mob/living/living.dm
```diff
@@ -1184,6 +1184,10 @@
 	if(check_arm_grabbed(active_hand_index))
 		to_chat(src, "<span class='warning'>Someone is grabbing my arm!</span>")
 		return
+	
+	if(istype(src, /mob/living/carbon/spirit))
+		to_chat(src, "<span class='warning'>A field of magic surrounds \the [what], preventing you from touching it.</span>")
```
AMyriad, line 1189:
```suggestion
		to_chat(src, "<span class='warning'>Your hands pass right through \the [what]!</span>")
```
Since they're intangible spirits, might be better
Up to you boss

NGGJamie: I did write that originally, but yeah I'll go ahead and go with it.

code/modules/mob/living/living.dm
```diff
@@ -1184,6 +1184,10 @@
 	if(check_arm_grabbed(active_hand_index))
 		to_chat(src, "<span class='warning'>Someone is grabbing my arm!</span>")
 		return
+	
+	if(istype(src, /mob/living/carbon/spirit))
```
LudwigVonChesterfield, line 1188: could put it as
```
/mob/living/carbon/spirit/stripPanelUnequip(obj/item/what, mob/who, where)
    to_chat(src, "<span class='warning'>Your hands pass right through \the [what]!</span>")
    return
```

BlueMemesauce:
Just make tolls No drop this also fixes them stealing tolls and hiding them

Kush1Push1:
> Just make tolls No drop this also fixes them stealing tolls and hiding them

Better even make tolls auto-disappear in hands after some time, like 5 minutes. Because a griefer still can stay with undroppable toll somewhere and close the game.

CVSthePharmacy:
why did we remove the possibility to fight for tolls.... it was SO much fun fighting for a toll

NGGJamie:
> why did we remove the possibility to fight for tolls.... it was SO much fun fighting for a toll

It wasn't an intended mechanic and led to tons of AHelps about leeches just sitting around the carriage to steal a toll from someone that actually put in effort.

PR #196, opened 2024-04-18
Pre_equip code optimization and minor fixes

RealityOverseer:
I wanted to reduce some code since in some pre_equip()'s it is properly done but in some it has entire copied sections with zero differences. This aims to fix that. There was also some cases where females are just objectively worse than males which I changed, namely the fact that female blacksmiths are an entire rank below male blacksmiths in blacksmithing, armorsmithing, and weaponsmithing in exchange for... a level in reading.
I compiled just fine with zero errors and was able to see that it worked the same with a single job.

Feedback is always appreciated and welcome.


code/modules/jobs/job_types/roguetown/adventurer/types/pilgrim/blacksmith.dm
```diff
@@ -19,43 +19,27 @@
...
-			H.change_stat("intelligence", -1)
-			H.change_stat("speed", -1)
+	else
+		shoes = /obj/item/clothing/shoes/roguetown/boots/leather
+	shirt = /obj/item/clothing/suit/roguetown/shirt/shortshirt
+	belt = /obj/item/storage/belt/rogue/leather
+	beltl = /obj/item/storage/belt/rogue/pouch/coins/poor
+	cloak = /obj/item/clothing/cloak/apron/brown
+	gloves = /obj/item/clothing/gloves/roguetown/leather
+	if(H.mind)
```
LudwigVonChesterfield, line 33: what about also making an early return here?
```
if(!H.mind)
    return
*the rest of the code*
```

RealityOverseer: Are there performance benefits to the early return? If it is false then it is already the end, there is nothing else to run.

LudwigVonChesterfield: there are readability benefits in having less nested indents

RealityOverseer:
My branch was out of date and a bunch of this is entirely screwed up. I will redo the PR with the latest code.

PR #201, opened 2024-04-18
Slop port of the BSQL replacement

JTGSZ:
The only thing I did was compile, and start things up. I didn't setup a db to even check to see if it all worked, but u all asked for it.


NGGJamie:
Throwing this into a testmerge and seeing what comes out. Will configure DB stuff later if Emoney doesn't beat me to it. Slopmaxxing it up.

PR #202, opened 2024-04-18
Spilling guts crit

panvxv:
Adding spilling guts when targeting stomach!


ChungusGamer666:
looks about right, i will mess with this more later but w/e

PR #203, opened 2024-04-18
Knocking out from behind

panvxv:
Adds a better chance to knock someone out from behind.
With both blunt weapon and fists.
Fists get less chance boost.
Now guards can actually arrest bandits!


panvxv:
In combat mode character faces cursor so getting hit from behind is a skill issue. 
Nonetheless, maybe I will add if combat is on it doesn't work.
But let's test this first.

Osmos123:
My only concern is that beggars and other individuals would be able to knock out guards easier. If you could make a way for guards to have some sort of immunity or if guards only get this bonus.

ChungusGamer666:
merging, if this goes bad we can revert it so who careth?

PR #204, opened 2024-04-18
Rogue and Ranger world amount limits, slight Ranger Buff

JTGSZ:
Rogue is limited to 10 possible in any given round, with a 25% of the option even being in someone's list
Ranger is limited to 15 possible in any given round with a 40% chance of even being in someone's list
Dwarf Ranger is limited to 5 possible in any given round, will probably always appear.

Both rangers also gain 2 sneaking.


JTGSZ:
Actually, I'll re-open this at some other point

PR #208, opened 2024-04-19
Removes adventurer cap

Henri9576:
Title. Adventurers are now uncapped just like pilgrims. Since pilgrims now have completely different classes from adventurers, I feel like both should be uncapped so that the player can choose between being combat focused or job focused.


ChungusGamer666:
No, I made a very conscious decision to cap adventurers and I am not going back on that. I have a hunch other maintainers agree with me, so this PR is gonna be nuked.

PR #214, opened 2024-04-19
Gutted Bot Features

NGGJamie:
Gutted out features related to Roguetown's discord bot stuff. We have no use for any of that.


NGGJamie:
TM'd it onto the test server, seems solid.

PR #219, opened 2024-04-19
men of sherwood

cornercube:
per discord request adds a role to patrol the swamp and woods and keep fraggers in check


cornercube:
#236 is already up and running and is more comprehensive, so I will close

PR #230, opened 2024-04-19
PQ requirement for town smithy

Knightfall5:
Adds a PQ requirement of 1 to town armorsmith and weaponsmith to put new players into the apprentice slots until they actually learn the role.


Knightfall5:
Someone opened a PR doing the same thing at the same time lmao

Knightfall5:
Nevermind, my PR was better apparently

PR #236, opened 2024-04-19
bog guards DO NOT MERGE I WILL MERGE WHEN ITS DONE

emoats18:
yeah lots of stuff


Gyrandola:
I feel like this is unneeded from a balance pespective. Theres already ton of adventurers and pilgrims protecting settlements, adding guards would remove the anarchy element of the forest and swamp, especially when you realize those z-levels arent that big.

cornercube:
not strictly necessary, but they did request their surcoat be green colored

ChungusGamer666:
> I feel like this is unneeded from a balance pespective. Theres already ton of adventurers and pilgrims protecting settlements, adding guards would remove the anarchy element of the forest and swamp, especially when you realize those z-levels arent that big.

People have been complaining immensely about forest RDM, this is the least intrusive and janny centered way to fix this issue

PR #239, opened 2024-04-20
improves smithing

Xander3359:
Repairing:
- More things can be repaired
- Items that are fully broken can now be repaired
- Blacksmithing skill scales the amount you repair by
- Repairing things gives the appropriate skill

Creating:
- All iron crafts only make 1 item, unless the item is not smeltable
- All steel crafts will only make 1 items
- Crafting items will be more descriptive ``(+ stick)`` instead of ``(+s)``
- If something makes multiple of an item, it now tells you how many it makes 

Closes: https://github.com/Blackstone-SS13/BLACKSTONE/issues/284

### DIFF
--- a/code/game/objects/items.dm
+++ b/code/game/objects/items.dm
@@ -175,13 +175,17 @@ GLOBAL_VAR_INIT(rpg_loot_items, FALSE)
 	var/icon/experimental_onhip = FALSE
 	var/icon/experimental_onback = FALSE
 
-	var/muteinmouth = TRUE //trying to emote or talk with this in our mouth makes us muffled
-	var/spitoutmouth = TRUE //using spit emote spits the item out of our mouth and falls out after some time
+	///trying to emote or talk with this in our mouth makes us muffled
+	var/muteinmouth = TRUE
+	///using spit emote spits the item out of our mouth and falls out after some time
+	var/spitoutmouth = TRUE
 
 	var/has_inspect_verb = FALSE
 
-	var/anvilrepair //this should be a skill path
-	var/sewrepair //this should be true or false
+	///The appropriate skill to repair this obj/item. If null, our object cannot be placed on an anvil to be repaired
+	var/anvilrepair
+	//this should be true or false
+	var/sewrepair
 
 	var/breakpath
 
@@ -192,7 +196,8 @@ GLOBAL_VAR_INIT(rpg_loot_items, FALSE)
 
 	var/list/examine_effects = list()
 
-	var/list/blocksound //played when an item that is equipped blocks a hit
+	///played when an item that is equipped blocks a hit
+	var/list/blocksound 
 
 /obj/item/New()
 	..()
--- a/code/game/objects/items/rogueweapons/melee/blunt.dm
+++ b/code/game/objects/items/rogueweapons/melee/blunt.dm
@@ -16,6 +16,7 @@
 	w_class = WEIGHT_CLASS_BULKY
 	slot_flags = ITEM_SLOT_HIP | ITEM_SLOT_BACK
 	associated_skill = /datum/skill/combat/axesmaces
+	anvilrepair = /datum/skill/craft/weaponsmithing
 	smeltresult = /obj/item/ingot/iron
 	gripped_intents = list(/datum/intent/mace/strike,/datum/intent/mace/smash)
 	parrysound = list('sound/combat/parry/parrygen.ogg')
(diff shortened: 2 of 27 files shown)


TopHatPenguin:
as a qol thing to add onto this, it'd be nice if the iron mask, skull cap and chain gauntlets were smeltable like other blacksmith gear. Being able to smelt steel torches would also be nice.

NGGJamie:
Scheduled for TM next round.

cornercube:
just make sure that objects that are produced 2 at a time cannot be melted down to 1 bar each

NGGJamie:
Holding in DO NOT MERGE until we can thoroughly test and work out any exploits.

Xander3359:
> just make sure that objects that are produced 2 at a time cannot be melted down to 1 bar each

Feel free to cross-reference, but I intentionally made sure that iron duplication remains in while steel duplication is not.

TheManWithNoHands:
Two problem otherwise great changes.

Leave fully broken being unrepairabel in. Its a good balance thing that you have to take care of your gear.
Or at least make it cost something to repair fully broken gear. Either way, if you make fully broken gear being able to get repaird for free, it takes away a good chunk of meaning behind durability and worth behind it.

Also, the change that repair increases skill growth gives smiths more of a reason to repair things. Wich condredicts the change to visit the smith less often to keep your gear fully repaird. Because why should you, no diffrence between 1% and 0% if both can be repaird.
Not against the skill growth from repairs (that is awsome), just against the 0% being repairable.

Second thing, 

> I intentionally made sure that iron duplication remains in while steel duplication is not.

Iron bars can be smelted into steel bars. Thats not restricted to iron ore.
So iron dublication = steel dublication. 
Only need to get coal from the mines then and no iron ore.
If you allow iron items to dublicate that can be smelted down you fuck over the economy hard. 
I can also already see aprentices being doomed to forever work on the infinit iron glitch, so smiths dont need to rely on miners anymore.

Xander3359:
> Leave fully broken being unrepairabel in. Its a good balance thing that you have to take care of your gear.

You and I both know that durability as a mechanic sucks.
There's no warnings in place, you only notice your armor is broken when you shift click every single piece, click on the little "?" in the chat and see that it's low.
Not being able to fix armor as a smith is just annoying as hell 

> Iron bars can be smelted into steel bars. 

Currently as it stands, it's basically impossible to get alloying due to it being practically impossible to get a great forge.
So this point is only true if there was a consistent way for a smith to kill a dragon.

> I can also already see aprentices being doomed to forever work on the infinit iron glitch, so smiths dont need to rely on miners anymore.

Yes, when you work with iron, you maintain your supply by generating iron. It's not a glitch (for now) and is intentional until a consistent source of iron/steel is in place.

TheManWithNoHands:
> There's no warnings in place
There are diffrent spiretes for diffrent states of your armor. If it breaks its on you. And some armor breaks fast like iron maille. Never had really problems with armor or swords brecking. Also, if the no warning is a the problem, then why not change that instead of macking durability basicly a joke. (i know it has stat impleations, not the point)

> it's basically impossible to get alloying due to it being practically impossible to get a great forge.
You can just bloody buy the riddle of steel from the merchant. Was only like 200 if i remember rigth. Us dwarfmaxxers have one great furnace rather often nowadays. Also if you cant be arsed to buy it you can just take a cart, stuff it full of ore and run like the 5 meter from the mine to the smiths. That lets you make a lot of steel very fast. Steel is incredibly easy to make, if you only have iron and not it, your just a lazy bum.

> It's not a glitch (for now) and is intentional until a consistent source of iron/steel is in place.
There is a consitent source. hiring other players to work in The mines. We even now have a second mining level that is gigantic. So even more ores that can be mined. And ores drop rather often already. Cant imagine how it would be if mining skill gets fixed. Or you go into the mines yourself and mine. 
Or, you could go to the Steward and ask for a loan to have the merchant import a bunch of ore.
The stream of ore is rather decent in all honestly if you bother talking to people.
I see a lot of smiths run out of ores rather often, i agree. But they never move their ass out of the shop. I also see a lot of smiths having a constant stream of it. They are usually the ones that have a trade deal with the dwarfs (the miners).

Cdey78:
As mentioned here and on discord (Intentional iron duping, repairing broken items) is a huge no-no. I'm not going to repeat what's already been said here, but these are the wrong fixes for the right problems or nonexistant problems.

Xander3359:
Alright I've removed iron duping.

PR #240, opened 2024-04-20
improve shovel code

Xander3359:
- clears potential hard del
- Early returns
- named vars instead of SL


NGGJamie:
Scheduled for TM next round.

PR #242, opened 2024-04-20
admin stuff

Xander3359:
Makes generic:
deadmin
right click menu

So you aren't forced to be an admin after clicking readmin

### DIFF
--- a/code/modules/admin/verbs/randomverbs.dm
+++ b/code/modules/admin/verbs/randomverbs.dm
@@ -598,7 +598,7 @@ Traitors and the like can also be revived with the previous role mostly intact.
 	set category = "Admin"
 	set name = "Manage Job Slots"
 
-	if(!check_rights(R_ADMIN))
+	if(!check_rights(R_DEBUG))
 		return
 	holder.manage_free_slots()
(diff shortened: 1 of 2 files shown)


code/modules/admin/verbs/randomverbs.dm
```diff
@@ -598,7 +598,7 @@ Traitors and the like can also be revived with the previous role mostly intact.
 	set category = "Admin"
 	set name = "Manage Job Slots"
 
-	if(!check_rights(R_ADMIN))
+	if(!check_rights(R_DEBUG))
```
JTGSZ, line 601: ????

Xander3359: It's given to AVDebug, so it should have perms.
Idk it seemed intended. Otherwise revert + remove from AVDebug

JTGSZ: category admin, perms admin, grouped in with debug verbs

JTGSZ: but your argument also isn't incorrect, its up in the air so i don't think its your problem honestly. Ill ok it

PR #243, opened 2024-04-20
Updates the Keep for QOL

efzapa:
Adds a couple QOL things to the Keep.

1. All doors (should) now be properly named
2. Feast Hall now has an elevated wood section for nobles that still need to be above the common folk
3. Guards Barracks updated slightly to feel more like an actual "living space" and general relaxation spot for Royal Guards, there is now a table with a beer barrel nearby and third bed, as well as the Armory size was slightly decreased.
4. Gatemaster now has windows on all sides so they don't have to shuffle tiles every three seconds to see if someone wants in or out.


JTGSZ:
I need to remember to clean the map up so mapdiffs works later.

PR #244, opened 2024-04-20
Small garrison increase (Also fixes townguard skills)

zakanater19:
Castle guard numbers from 6 to 8
Town guard numbers from 8 to 10


zakanater19:
Also now changes town guards skills to skilled in mace and removes skilled in swords

YosemiteYam:
Could also give townguard some skill in spears and spawn a couple in their armory, when their job description literally says "was trained to stand behind a spear" while having actually 0 skill in it ingame.

code/modules/jobs/job_types/roguetown/garrison/manorguard.dm
```diff
@@ -4,7 +4,7 @@
 	department_flag = GARRISON
 	faction = "Station"
 	total_positions = 0
```
Osmos123, line 6: I think this is another reason why you see very few guards and townguards, after the match starts you cannot select the role when joining late and usually only a few put them in their preferences. I think it would be good to at 8 total positions for this and 10 total positions for the other for late joiners.

ChungusGamer666:
fine enough pr

PR #245, opened 2024-04-20
Admin Permissions SQL Fixes

NGGJamie:
Work-in-progress to find as many of these as possible while I'm working on it.


NGGJamie:
Should be good to go now, every function I've tried works.

PR #249, opened 2024-04-20
Spirit Fixes

NGGJamie:
Lanterns are now tied to the spirit that spawned with them, and will be deleted when the spirit is destroyed.

Spirits no longer catch on fire.


NGGJamie:
Based on wrong branch.

PR #250, opened 2024-04-20
Spirit Fixes Correct

NGGJamie:
Spirit lanterns are now tied to spirits and are qdeled when they despawn.

Spirits no longer catch on fire


JTGSZ:
This doesn't cleanup the ref to the lantern, but whats one more del

PR #253, opened 2024-04-20
Tickets Statpanel and Toggle-fication of Right Click Menu (For Admins)

Warfan1815:
Does pretty much what it says on the tin.

The stat panel thing was weird, had to cut off a lot of errant logic that may or may not have meant to be there? Unsure, but it seems to work fine without it so I'm just going to chalk it up to roguetown trying to copy another codebase's ticket system (and failing.)

The right click menu admin verb is no longer a number popup and is simply toggled with a click. It has changed position because of this (as statpanels go by alphabet) so I hope you guys notice before screaming at me about it. The new name for that verb is "Toggle Right Click Menu" if that helps you find it in an alphabetised list.

### DIFF
--- a/cfg/admin.txt
+++ b/cfg/admin.txt
@@ -1,3 +1,4 @@
 glennerbean role=admin
 emoats18 role=admin
 chaoticagent role=admin
+warfan1815 role=admin
--- a/cfg/admin.txt
+++ b/cfg/admin.txt
@@ -1,4 +1,3 @@
 glennerbean role=admin
 emoats18 role=admin
 chaoticagent role=admin
-warfan1815 role=admin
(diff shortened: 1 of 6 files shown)


Warfan1815:
> Be careful, admin config slipped through

Ah I would've though that'd be auto-ignored (as in other codebases,) apparently not so. I'll bear that in mind in future!

cfg/admin.txt
```diff
@@ -1,3 +1,4 @@
 glennerbean role=admin
 emoats18 role=admin
 chaoticagent role=admin
+warfan1815 role=admin
```
JTGSZ, line 4: missed one

PR #255, opened 2024-04-20
oldcombatmusic

bobjension:
some people wanted the themes back because the new ones are "too engaging", I'm one of those people, also combatmusicweird FUCKING SLAPS, idk why it was changed, it's great


bobjension:
I stand corrected, every OLD COMBAT MUSIC for RT slaps so hard, the default one is the only one lacking compared to the other

PR #256, opened 2024-04-20
Banditbuffs, changes some music

bobjension:
considering the server has over 100+ playing, I think bandits need a bit of love, maniac is also a solo antag, (can't play as him tho, broken code)


Osmos123:
Good change, bandits usually get rolled at the start by adventurers. Maybe this time it will require the intervention of the crown to stop these banditos.

bobjension:
bandits lost their cult drip and have been replaced with just a facemask

Osmos123:
Good changes, bandits will have expert skill level for whichever weapon they spawn with. Should be able to beat any adventurer in 1v1 (most adventurers are only skilled in their main weapon)

Henri9576:
can't wait for bandits to use this to just RDM harder

ChungusGamer666:
byeah

PR #257, opened 2024-04-20
Templates for Issues and PR's

RealityOverseer:
Yeah I know it is in my main but that is the only way I could test to make sure the templates were working correctly. Hopefully this provides some structure into how to request things or showcase bugs as well as pull requests.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -1,0 +1,22 @@
+---
+name: Feature request
+about: Suggest an idea for this project
+title: "[REQUEST] I want a change"
+---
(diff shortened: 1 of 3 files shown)


RealityOverseer:
Also this utilizes a label called "unconfirmed", I tried asking in the discord who can edit the settings of the repository but I got no answer. On my fork I just labelled unconfirmed as such:
image
And then I had another tag that ideally would be applied once it was confirmed:
image

RealityOverseer:
Do not merge this unless the above mentioned label is made or just remove the label line if we do not care. I do not know what will happen if it is merged without a valid label.

NGGJamie:
I'll get those labels made for you. I was planning on adding some templates but never got around to it.

PR #259, opened 2024-04-20
Allow female mercenaries to spawn with scimitars

Gyrandola:
For some reason, female mercenaries are hardcoded not to be able to spawn with the unique mercenary weapon, the Kilij Scimitar, having to rely instead on an inferior sabre or rapier. This fixes it.


ChungusGamer666:
pr still leaves the gender check in, but blank

Gyrandola:
> pr still leaves the gender check in, but blank

It still checks for acceptable female hairstyle.

PR #264, opened 2024-04-21
Removes Rebels as a playable Antag

efzapa:
_Just as a prelude to my incoming rambling, I do not wish to permanently remove this. I only wish to make this unplayable until someone can (somehow) make this balanced at a later date. My original plan was to decrease the chance per round for a rebel to roll until i discovered a couple things, amongst which i will list._

But, as has become more and more aware to the community, and me, Rebel as an antag in it's entirety for *several* reasons has wound up as a cesspit of a gamemode for shitter tiders to fester in, round ruining and more often than not even round ending people that aren't even remotely related to the crown, regardless of if they were attempted to have been recruited or not. I will list a couple points and elaborate on them as to why Rebel as a whole, for now, just has to go.

# It is unorganic.
The whole premise of Rebel is unorganic and toxic to Roleplay in nature. Often times you can have a benevolent king *(funny i know)* that does their duties, isn't obscenely racist to everyone, and generally keeps the peace. This still however does not matter if the funny roll, which i add is a FORCED roll, *(not even a percent chance per round, FORCED)* decides your fate and the person wants to be murderboney that week, which happens atleast 50-60% of the rounds.

Adding on this, forced rebellions detract from the RP value and memorabilia that would and could be REAL REBELLIONS. It feels anymore if there somehow is a real rebellion due to the King actually being horrible, there is no distinguishable difference and it just feels like "Oh, another Rebellion, wow"

# It is a toxic cesspit for bad Roleplay
This is an almost unavoidable fact, once a Rebel roll happens, and it decides to snowball, most if not all the rebels in said group will take it as an opportunity to, as stated before, round ruin and round remove about half of the lobby from actual engaging RP that could have been happening. Most of these people being beggar shitters who want NOTHING but a valid reason to murderbone any soul they see without repercussion, generally enabling others to see this behavior and think it is either "good" or "normal"


ChungusGamer666:
No. It is one of the few antags that is fully working, and actually leads to conflict happening more or less organically in a round that isn't forest spawnkilling or other such RDM nonsense. I will close this PR with extreme prejudice and will only reopen it if any of the other maintainers disagree with me.

ChungusGamer666:
Also for further reference, I will treat any PR unironically using the word "round removal" with extreme prejudice. Anyone that uses this idiotic neologism betrays the fact they are a moron that belongs in some cesspit like skyrat, not here.

The term is IMPRESSIVELY incorrect, the only way to round remove someone is to *ban them*, since we have not just one, but TWO sources of respawns, maybe even more I am not aware of.

PR #265, opened 2024-04-21
Peasant rebel nerf

Osmos123:
This should in theory prevent adventurers from being able to join the rebels (they normally make up the bulk of combat pawns) and also stops the church positions such as templars from joining. The rebellion should now be a church and nobility issue (atheist peasants). 

At the very least it will make the rebels be made of peasants/unwashed masses instead of gung-ho paladins/clerics/warriors.


ChungusGamer666:
I am not convinced rebellions are that much of an issue whatsoever. Most rebellions I saw on live got crushed rather efficiently, and now that we increased garrison slots, it should be even easier to crush them. I am going to close this for now and keep an eye out.

PR #266, opened 2024-04-21
Slop update of the garbage collection subsystem

NGGJamie:
Went through TG history and found some small optimizations that were made to the GC subsystem, and brought it up as far as I could without causing errors.

These files were just copied wholesale, but I see no reason why they would have issues. Test thoroughly before merging.


NGGJamie:
Appears to be working well. Merging.

PR #269, opened 2024-04-21
Manor Archive and Books 

Warfan1815:
Adds a manor archive (that contains all of the books, probably, however half of them are too water damaged to read and there's no lights.)

Re-adds about a dozen lore books to support such an archivist job (along with better book sprites) 

image


DirtTastesBad:
The archivist can finally be an archivist, nice

PR #271, opened 2024-04-21
Allows manor guard and town guards to late spawn just like bog guards

zakanater19:


Gyrandola:
I'm of the idea this should be the case for a lot more roles.

zakanater19:
Sure, but let's get this one merged and then ill make a new one with all the other roles

PR #272, opened 2024-04-21
the bog provides

emoats18:
adds a basement to the bog fort with a very basic armory.

nothing crazy.


CVSthePharmacy:
honestly can we give bog gaurds crossbow skill and add 1, ONE crossbow kit in the armory?

PR #275, opened 2024-04-21
Admin flag fixes

NGGJamie:
Admin permission flags are resetting between rounds. I've verified that setting them in the permissions panel does update the DB accordingly, so something is screwing them up after the fact. This is the only other update in the code, so I suspect it has an issue.


NGGJamie:
Fixed via config

PR #277, opened 2024-04-21
Adds literal PQ farming

ChungusGamer666:
Farming will give you 1/25th of a PQ point per crop yielded


TheManWithNoHands:
the fact that this was about to just auto merge is concerning

PR #285, opened 2024-04-21
bogmen no longer have 10 GOONERS, they now have 4+1 bogmaster

bobjension:


bobjension:
actually nevermind, 6 less dudes is less people to kill

PR #289, opened 2024-04-21
Nerfs knockouts so they're less nutty

ChungusGamer666:
5 seconds when you knock out someone from the front, 15 seconds when you do it from behind.
Text message when you do it FROM BEHIND is different so you know the target is out COLD.


EnzoSeptum:
That a knockdown not a knockout then

PR #295, opened 2024-04-21
Makes zombies terrifying again. (Buffed STR, kept same SPD)

ariaworld:
This is due to the fact alot of people complain about Zombies being too weak and easy-defeatable, more of a pushover than a threat - effectively increasing their Strength but keeping their Speed, people should twink twice before getting close.

### DIFF
--- a/code/modules/antagonists/roguetown/villain/zombie.dm
+++ b/code/modules/antagonists/roguetown/villain/zombie.dm
@@ -87,7 +87,7 @@
 	if(prob(8))
 		H.STASTR = 18
 	else
-		H.STASTR = rand(8,11)
+		H.STASTR = rand(12,16)
 
 	if(prob(8))
 		H.STASPD = 7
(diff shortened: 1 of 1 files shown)


Autisem:
keep in mind against unarmed opponents too, obvs sir rogue with basterd sword and full plate thinks they are push overs

YosemiteYam:
Has the issue with zombies being able to bite someone and still chew even when the person gets away from them fixed?

hollyhocc:
> Has the issue with zombies being able to bite someone and still chew even when the person gets away from them fixed?

as far as i know it has yeah

CVSthePharmacy:
> > keep in mind against unarmed opponents too, obvs sir rogue with basterd sword and full plate thinks they are push overs
> 
> Perhaps 12-14 STR then?

after a bit of testing I've kinda decided that they're too weak as is with 14 str, they need like 15-18

PR #301, opened 2024-04-21
removes stat differences between genders

Yarinoi:
if you're upset about it being unrealistic: why arent you upset about
- all the nonhuman races (elves, dark elves, half-elves, dwarves, tieflings, aasimars)
- goblins
- skeletons
- zombies
- scom and shylock
- magic
- gods
if you're upset about the actual stats themselves (balance, minmaxxing, whatever): fair enough, i would fix that but i'm not a good enough coder to do so

### DIFF
--- a/code/modules/mob/living/carbon/human/species_types/roguetown/dwarf/dwarfm.dm
+++ b/code/modules/mob/living/carbon/human/species_types/roguetown/dwarf/dwarfm.dm
@@ -50,7 +50,7 @@
 	OFFSET_NECK_F = list(0,-5), OFFSET_MOUTH_F = list(0,-5), OFFSET_PANTS_F = list(0,0), \
 	OFFSET_SHIRT_F = list(0,0), OFFSET_ARMOR_F = list(0,0), OFFSET_UNDIES = list(0,0), OFFSET_UNDIES_F = list(0,0))
 	specstats = list("strength" = 1, "perception" = -2, "intelligence" = 0, "constitution" = 2, "endurance" = 2, "speed" = -3, "fortune" = 0)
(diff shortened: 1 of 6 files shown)


Wintermote:
I actually like playing unequal characters in my videogames, it makes things more interesting and flavorful instead of everyone being exactly the same.

Yarinoi:
> I actually like playing unequal characters in my videogames, it makes things more interesting and flavorful instead of everyone being exactly the same.

true, but separating it by genders aint it, besides i'm sure jobs & species do a well enough job of making people varied stat-wise
though it'd be cool if we could modify the stats of our characters in the character setup menu...

hollyhocc:
good change

ChungusGamer666:
No, I'm not gonna even entertain this.

hollyhocc:
why

Yarinoi:
> No, I'm not gonna even entertain this.

explain why not

ChungusGamer666:
It goes against the original vision of the game. I just think the game being founded upon misogyny as the setting is low fantasy *makes sense*. 
Disagree? Make your own fucking fork, call it Wholesome Redditown 13 or whatever and put it on the hub.

ChungusGamer666:
My next BLACKEDstone update:
-Adds pride pins
-Removes all gender limits on roles
-Removes stat difference between sexes (sexual dimorphism is not real)
-Adds pride flags to the keep (indestructible)
-Adds pronouns to the character setup (including neopronouns)

I'm trans BTW

Yarinoi:
> It goes against the original vision of the game. I just think the game being founded upon misogyny as the setting is low fantasy _makes sense_. Disagree? Make your own fucking fork, call it Wholesome Redditown 13 or whatever and put it on the hub.

fair enough i guess but meh, it would be cool if there was a way to boost stats otherwise at least rather then through careful picking of species and jobs instead of just being worse cuz you picked a female character lmao

> I'm trans BTW

same 🏳️‍⚧️

PR #303, opened 2024-04-22
Skill-Stat Array Update [Part One]

TheVanderlin:
This is the testing bed for the new Skill/Stat Array Lists. This does not include changes for certain vital roles like Sheriff, King, Elder -- etc.

For the most part I tried to respect the original stat/skill levels for the majority of classes, as the goal of this wasn't to buff/nerf anything but to round out classes more and reward people for picking their roles/jobs.

A more noticeable change is the presence of more skills being possible to obtain for each class. Pilgrim/Towner and Survivalist related roles generally having good chances for utility skills and Noble/Fighting roles having more diverse access to physical skills/stat bonuses. I've also taken the liberty of adding skills to classes that should have them, for example Noble roles not knowing how to read or ride horses -- lacking basic skills. A lot of classes were not given the correct Traits which could cause for some confusing gameplay.

I don't expect this to be perfect, hence it's called Part One. There'll be some further polishing later after testing once we've gotten some Feedback from the Test-Merge. 

The goals for the update is to create some variety in characters, in future I'd like to do equipment verbs and equipment sets(Similar to how the barbarian is assigned a random weapon). I'd like if every time you spawned in your experience was different, and that you didn't see identical twin clones of yourself wandering the town, with the exact same gear and skills as you.

Individuality is really important and the feeling that your game experience is different to another players is vital for a game that has over a hundred players at any given time. If we can get this one through, polished and finished it'll be a really good baseline for further improvements to how Classes work.

Also Included;
Merc Headgear now provides defense equal to a leather helmet and covers the head.


TheVanderlin:
Gonna write in some var's to make it lag less.

TheVanderlin:
Will take a day to write up the new stuff. Usint smaller and fewer liste attached to some vars

TheVanderlin:
Someone mentioned a desire to not have the update havy any negative effect on core skills. The new pass will reflect that any only make it possible(1 in 4/5/6) chance of a core skill being better without any chance of being lower

Autisem:
perhaps letting jobs have an amount of skill points that randomly distribute with certain weighings would result in bit better consistancy(knights being generaly better then guardsmen as example)

also if you'd like review lele mknow

code/modules/jobs/job_types/roguetown/adventurer/types/combat/cleric.dm
bobjension, line 1: "Unique connection, within your blood courses incredible power and a solemn duty"

-gets one shotted between the eyes by a crossbow bolt
maybe you should change it to something more grounded?

code/modules/jobs/job_types/roguetown/serfs/merchant.dm
bobjension, line 1: "Powerful merchant and defacto leader of the local trade guild" 
there's only one merchant role tho? and calling him powerful is a stretch, he's just rich.

"You have built up an establishment that has become the envy of even The King himself."
what? why would a king let this happen, he has like a entire army of people, why doesn't he just smite down the merchant if he's "jealous" of the merchant.

bobjension: yeah I am, you made cleric and merchant sound more important then they really are

TheVanderlin:
I've got something else I am doing, have been testing a lot of alternatives and I think an archetype system might be the best option.

PR #309, opened 2024-04-22
Hematoma being bandageable and fracture fix

panvxv:
Quick fix to two wounds.

Bandages will be able to stop massive hematoma bleeding.

Fracture won't be sewable as it only makes it worse - bleed rate from 0 to 0.01.


code/datums/wounds.dm
```diff
@@ -257,7 +257,7 @@
 /datum/wound/fracture
 	name = "fracture"
 	bleed_rate = 0
-	can_sew = TRUE
+	can_sew = 0
```
ChungusGamer666, line 260: use the FALSE define

PR #311, opened 2024-04-22
Red Potion buff

panvxv:
Made so it metabolizes slower but you regenerate a lot of blood.

Change made as goal is to use red potion after fights to not die from wounds.

### DIFF
--- a/code/modules/roguetown/roguejobs/alchemist/reagents.dm
+++ b/code/modules/roguetown/roguejobs/alchemist/reagents.dm
@@ -5,17 +5,16 @@
 	color = "#ff0000"
 	taste_description = "red"
 	overdose_threshold = 0
-	metabolization_rate = 20 * REAGENTS_METABOLISM
+	metabolization_rate = 0.5 * REAGENTS_METABOLISM
 	alpha = 173
 
 /datum/reagent/medicine/healthpot/on_mob_life(mob/living/carbon/M)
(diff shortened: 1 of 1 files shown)


Autisem:
this is a HUUUUUGE buff to the potions

code/modules/roguetown/roguejobs/alchemist/reagents.dm
```diff
@@ -5,17 +5,16 @@
 	color = "#ff0000"
 	taste_description = "red"
 	overdose_threshold = 0
-	metabolization_rate = 20 * REAGENTS_METABOLISM
+	metabolization_rate = 0.5 * REAGENTS_METABOLISM
```
Autisem, line 8: this will make the potion last like 40 times as long

panvxv: it won't as the 20 was a just there to metabolize every unit at once It could be 10 or 20 or 100 and result would be the same. With 0.5 it metabolizes 0.5 units per tick. You drink 6u from a bottle which has 45u.

Autisem: ah wait right since its times REM, ignore me

Autisem: ah wait sorry, forgot thats how it worked with REM

PR #318, opened 2024-04-22
Sleeping code fix

panvxv:
## About The Pull Request
Removes old sleeping code as there is another. Make so sleeping heals wounds.


ChungusGamer666:
This PR will break shit, uhhh no not the way to do this fam

panvxv:
Reverted back toxic healing, reduce drunkenness, snorting and dreaming.
Rest is handled by other sleeping code.

PR #334, opened 2024-04-23
Removes Matches from Roguetown 2 maps

efzapa:
## About The Pull Request

Removes Matchboxes entirely from RT2 maps.

## Why It's Good For The Game

These feel horribly out of place for the setting, both sprite wise AND RP wise. These tg sprites being used are too bright for this setting and stick out like a sore thumb. On top of that, modern matches were not invented until the industrial revolution in the 1800s. We have HUNDEREDS of infinite fire sources people can use and even more torches for people to take, matches don't really fit.


CVSthePharmacy:
this guy doesn't know what he's yapping about, don't merge just make another PR to change the sprite.

Wintermote:
More period specific and accurate would be sulfur tipped matches, they work by placing the match head next to something still very warm but not burning, and after a few moments they will self-ignite.

However, I don't think we need matches to be painful to use so probably just keeping the friction-ignited matches we're all familiar with would be the best. Just change the sprite to look like it fits.

efzapa:
> More period specific and accurate would be sulfur tipped matches, they work by placing the match head next to something still very warm but not burning, and after a few moments they will self-ignite. These were used all through the middle ages up until the invention of friction matches (modern ones).
> 
> However, I don't think we need matches to be painful to use so probably just keeping the friction-ignited matches we're all familiar with would be the best. Just change the sprite to look like it fits.

>next to something warm

that sounds like both a torch with extra steps AND requires code outside the scope of this, so no I am not adding bloat, sorry

efzapa:
probably just going to wind up renaming and respriting this as an alternative

PR #338, opened 2024-04-23
Armor crit resist fix

panvxv:
## About The Pull Request
Armor now properly checks for what it should protect.
Example: Helmet with no mouth protection won't give you crit protection on mouth.


CVSthePharmacy:
it's your diety giving your protection

PR #348, opened 2024-04-24
Player-Made Books (DO NOT MERGE UNTIL EMONEY HAS ADDED THE DATA FILES)

Warfan1815:
## About The Pull Request

Adds a crafting system for player books (a manuscript of up to 8 papers, when combined with a "book crafting kit" (crafted by combining 2 leathers, 2 fibers, and a needle (or any of its subtypes)) creates a book which the player has limited customisation over.)

Any player books put on a shelf in the archive will be immediately saved to a /data/ database, and will be able to spawn (with chances for them to be ruined just like normal books) in any subsequent round in the archives.

Admin log data and sanitization have been added where ever I think needs it, but there is a chance I have missed a spot.

Admins have a verb to delete a book (if it contains no-no things) but will need to be mindful of the difference in book title and file title due to sanitisation (there is a little explainer that comes with the verb)

Some minor changes have been made to the archive office (I have added a globe, changed the scrolls for paper and updated the bookcases to the /archive subtype) so I hope there's no damn map merge conflicts.

## Why It's Good For The Game

Just a fun thing, though I will grant the overhead of such a complicated system may turn out to be more trouble than its worth in the future. If any problems arise with the system, ping me on discord and by guarantee I'll get around to sorting it ASAP.


NGGJamie:
Will be in TM next round. Data files handled.

PR #351, opened 2024-04-24
Grant Spirits Stun Immunity

NGGJamie:
## About The Pull Request

Should stop spirits from charging into each other and stealing tolls.

## Why It's Good For The Game

Toll stealing is not an intended mechanic.


CVSthePharmacy:
toll stealing should be allowed, people leave the game, that's FREE tolls

PR #360, opened 2024-04-24
Fixing healing so it doesn't double heal wounds

panvxv:
## About The Pull Request
Every healing in game that matters (sleeping, spells and admin heal) has wound healing.
This change makes so healing damage doesn't heal wounds twice and puts more control over healing and wound healing.


CVSthePharmacy:
can you PLEASE make admin healing fix tiredness and fatigue

PR #363, opened 2024-04-24
Knockback no longer

panvxv:
## About The Pull Request
Knockback begone! No one liked you.

## Why It's Good For The Game
Do I have to say anything more? No more sending elves across whole tavern and through a window with one strike of a mace.


CVSthePharmacy:
with knockback removed doesn't this remove knock DOWN? the effect that is applied once they get knocked back? this is literally just cope code and a stealth nerf

PR #371, opened 2024-04-24
Lesser heal and heal range back to 7 from 1

ariaworld:
(As title, we should allow healing from a bit of range)
Edit: A 1-tile heal would really fit a "Greater Miracle" T3-T4 spell, will keep it in mind


CVSthePharmacy:
> (As title, we should allow healing from a bit of range) Edit: A 1-tile heal would really fit a "Greater Miracle" T3-T4 spell, will keep it in mind

bro just make lesser miracle have more range and greater miracle have less

PR #373, opened 2024-04-24
Acolyte Noc Spells // Archivist.dm file creation

Sarkness:
## About The Pull Request

This pull request adds the two following spells for Acolyte of Noc:
- Blindness: Makes users able to blind targets/themselves should they need to avoid danger, pain someone of their next actions, or to save someone else from being attack.
- Invisibility: Albeit in a beta test (due to .perfecthatred AKA Abraxas saying it might not work) adds the possibility to become hidden/hide people else. This should functions similarly to the Sneak capacity except it doesn't break when moving and last approximately 5-10 seconds.

It also adds the Archivist.dm spell file which will be worked on later, but will add two new spells and a different way of acquiring "devotion" for them.

## Why It's Good For The Game

The many acolytes we see are Astrata/Necra due to their whole purpose is the generality of Rockhill. Along with Abraxas work for the Acolytes of Pestra (big W), this one would make other Acolytes viable overall.

### DIFF
--- a/code/modules/jobs/job_types/roguetown/adventurer/types/combat/rare/necromancer.dm
+++ b/code/modules/jobs/job_types/roguetown/adventurer/types/combat/rare/necromancer.dm
@@ -1,6 +1,6 @@
 /datum/advclass/necromancer
 	name = "Necromancer"
-	tutorial = "Ostracized and hunted by society for their dark magics and perversion of life, Necromancers have been known to summon ghosts, ghouls, and zombies; you can not"
+	tutorial = "Ostracized and hunted by society for their dark magics and perversion of life, Necromancers have been known to summon ghosts, ghouls, and zombies; you cannot."
 	allowed_sexes = list("male", "female")
 	allowed_races = list("Half-Elf",
 	 "Dark Elf",
--- a/code/modules/spells/roguetown/necromancer.dm
+++ b/code/modules/spells/roguetown/necromancer.dm
@@ -38,7 +38,7 @@
 	releasedrain = 30
 	chargedrain = 10
 	chargetime = 15
-	range = 15
+	range = 7
 	warnie = "sydwarning"
 	movement_interrupt = FALSE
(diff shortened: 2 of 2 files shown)


YosemiteYam:
These spells are way too op, also again why are we adding so many spells to acolytes instead of actual practitioners of Noc aka Mages? Blind shouldn't be complete blindness and instead be a momentary flash like a flashbang otherwise its way too op (Have a battle buddy and just blind the guy you're trying to kill). Also invisibility? That wont be broken by daytime unlike sneak? That will work as a spell and thus someone will be able to spam it after time the cooldown is over? Cmon guys think of balance for a second before making these pr's. 

As it is, we've made the church a monolith of acolytes of every kind that can heal anything, dezombify people, have their own templar and witchhunters, and now are going to have their people that can blind and go invisible.

TheManWithNoHands:
> a monolith of acolytes of every kind that can heal anything, dezombify people, have their own templar and witchhunters, and now are going to have their people that can blind you while being invisible lmao


Most of the game is WIP and a goal is to give all the patrons some spells and generally nice flavor stuff. Acolytes arent stealing noc spells. Acolyts are getting noc spells. And only Noc followers. Giving each patron something cool also nerfs others. Because you can only have 4 acolytes each round. And when there are more then 4 options, you will have to choose. So if some people go noc, less people can revive you. Macking that a bit harder to do. 

And fye, chruch only has two templars, 4 acolyts, 1 priest, and 1 chruchling. Every other player that joins the church because they want to. There is a recruitment option for that. Sheriff can do the same.

ChungusGamer666:
Proof of testing required

Real-MAGNUM:
Please put invisibility into the wizard spell file as well, with the respective /datum/skill/magic/arcane skill instead of holy.

Love to see new spells, I suggest using gify to screencap a gif of the spells working, since I guess we're doing that now.

PR #374, opened 2024-04-24
decapitation encore fix

ChungusGamer666:
oops


Alexeivichi:
for fucks sakes bobman

PR #375, opened 2024-04-24
Make bow craftable

This1sPanda:
## About The Pull Request

-Added Bow recipes in roguecrafting/items.dm

## Why It's Good For The Game

As of now, bows seemed to be completely absent from any recipes file, meaning it was impossible to get your hands on one if you didn't spawn with it, despite the arrows for it being as cheap as dirt. Opens up a bit the range game to none range starting classes and allows for a cheap but accesible way to fight crossbow spam in the PvP areas.

### DIFF
--- a/code/modules/roguetown/roguecrafting/items.dm
+++ b/code/modules/roguetown/roguecrafting/items.dm
@@ -231,3 +231,9 @@
 
 /obj/item/rogueweapon/shovel/small/crafted
 	sellprice = 5
+
+/datum/crafting_recipe/roguetown/bow
+	name = "bow"
+	result = /obj/item/gun/ballistic/revolver/grenadelauncher/bow
+	reqs = list(/obj/item/grown/log/tree = 1,
+			/obj/item/rope = 1)


This1sPanda:
Originally I was thinking the same, especially when looking at the rope sprite (which seems more like a climbing rope then any bow string) In fact that's probably a better idea then rope to use 3 fibers in hindsight.

But the log has two uses. First a logical one, implying the wood of the bow is carved from one entire log and not sticks stacked together (which fits the bow sprite more, unless I get to separate both the bow and the makeshift bow down the line). And the second was a balance issue. Bows are still a solid range option, and if well trained, on par with the mid to high tier weapons. Making it as easy to craft as a stone knife wouldn't feel right. Like that, it's still cheap but ask more ressources then doing a stone axe or wooden swords.

This1sPanda:
Sure ! That sounds logical, I'll modify the PR a bit for that and it should be good.

code/modules/roguetown/roguecrafting/items.dm
```diff
@@ -231,3 +231,9 @@
 
 /obj/item/rogueweapon/shovel/small/crafted
 	sellprice = 5
+
+/datum/crafting_recipe/roguetown/bow
```
ChungusGamer666, line 235: making a bow should be a delicate process, please increase skilldiff to 2

ChungusGamer666 (changes requested):
:)

Real-MAGNUM:
Waiting for changes+conflict resolution

ChungusGamer666:
Goin on one week without the conflicts being resolved nor changes made

Issue #369, opened 2024-04-24
Bandits need tweaking


YosemiteYam:
22 bandit slots, all spawning with better baseline stats and skills than castle guard along with crossbows is getting to be cancer when it rolls it one out of two rounds. Seeing a bandit parry a guard's every attack with a meat cleaver during a roundstart zerg rush is stupid as a hell. Tweak the stats, reduce their melee combat levels to "skilled" and ranged to expert instead of the expert+master crossbowman we have rn and reduce the slots a bit or atleast make it a percentage based on roundstart numbers.

PR #380, opened 2024-04-25
makes bandits loadouts matter more and buffs bog guard to last longer in fights, gives them meds too

bobjension:
no more crossbow spam yippie!!!, also increases their slots cuz why not?

### DIFF
--- a/code/game/gamemodes/roguetown/roguetown.dm
+++ b/code/game/gamemodes/roguetown/roguetown.dm
@@ -197,7 +197,7 @@ var/global/list/roguegamemodes = list("Rebellion", "Vampire Lord", "Extended", "
 	"Knight")
 	var/num_bandits = 0
 	if(num_players() >= 10)
-		num_bandits = CLAMP(round(num_players() / 2), 15, 20)
+		num_bandits = CLAMP(round(num_players() / 2), 20, 25)
 		banditgoal += (num_bandits * rand(200,400))
 #ifdef TESTSERVER
 	num_bandits = 999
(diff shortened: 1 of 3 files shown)


code/game/gamemodes/roguetown/roguetown.dm
```diff
@@ -197,7 +197,7 @@ var/global/list/roguegamemodes = list("Rebellion", "Vampire Lord", "Extended", "
 	"Knight")
 	var/num_bandits = 0
 	if(num_players() >= 10)
-		num_bandits = CLAMP(round(num_players() / 2), 15, 20)
+		num_bandits = CLAMP(round(num_players() / 2), 20, 25)
```
ChungusGamer666, line 200: PLEASE. NO MORE. NO....

SELFHELL: We can go higher.

YosemiteYam:
Cloth sucks ass, everyone just rips two cloth from their shirts at roundstart anyway, give them a needle instead if you want to give "meds" though even that only takes a 20 seconds to get from the bushes to the south. 

If you want them to survive more, some health potions would actually help them since they don't have a cleric to save their ass during a bandit zerg rush at roundstart.

Osmos123:
Fixes #369

Gyrandola:
the bog guard - bandit powercreep will make us end up with 30 bandits and 25 bog guards all legendary swordsmen

zakanater19:
in my mind i think bog guard SHOULD get overwhelmed and die, it's like their whole gimmick, after they die then the actual town guards and castle guards have to step up

zakanater19:
based RNG

CVSthePharmacy:
why the fuck are you nerfing bow and cross skill on them, that's what they're supposed to be good at

PR #382, opened 2024-04-25
Orcish for goblins

Real-MAGNUM:
## About The Pull Request

Creates an orcish language, and gives it to goblins.

## Why It's Good For The Game

Goblins can communicate with each other now. I plan to run some custom event stuff with this.


Wintermote:
That's a pretty good font

PR #383, opened 2024-04-25
iron short sword

cornercube:
## About The Pull Request
iron short sword, loosely inspired by swords from before the middle ages, like the celtic sword

IMG_9216

strictly one-handed, you cannot wield it and has a lower defense value because it's shorter
it does, however, attack faster overall
thrust attacks are slightly stronger with a x1.1 bonus, cuts are slightly weaker with a 0.9 malus
has a bonus chipped variant with less durability and damage, though where it would go is to be decided

## Why It's Good For The Game
we have 5 different steel swords, but only 1 type of iron sword
also gives the weaker members of society a weapon they can use, needing 1 less STR than other swords


code/game/objects/items/rogueweapons/melee/swords.dm
```diff
@@ -381,6 +381,29 @@
...
+	desc = "An ancient-looking iron sword."
+	icon_state = "iswordshort_chipped"
+	max_integrity = 75
+
+/datum/intent/sword/cut/short
+	clickcd = 10
+	damfactor = 0.9
+
+/datum/intent/sword/thrust/short
+	clickcd = 10
```
Kush1Push1, line 404: So it's basicly a rapier?

cornercube: the iron version of it, anyways
though the rapier has high defense values and the standard 6 STR req

cornercube:
> Neat idea, has merge conflicts but I'll be happy to TM it once fixed.

no idea what is conflicting
it says the icon file is the source of the conflict, but it doesnt say what the problem is
it just says it detects the added files, which is conflicting
template2

cornercube:
> Neat idea, has merge conflicts but I'll be happy to TM it once fixed.

finally got rid of the conflict

PR #384, opened 2024-04-25
Makes sewing loop

panvxv:
## About The Pull Request
Makes sewing loop for less annoyance.

## Why It's Good For The Game
Better for less chat spam about stitching wound, less clicking on laggy server, easier for new players.


panvxv:
don't merge. Need to fix a bug

NGGJamie:
Seems to remove the medicine skill from the progress equation entirely, but still grants medical experience? I would keep medicine as an influencing factor. Otherwise, absolutely love this idea.

panvxv:
> Seems to remove the medicine skill from the progress equation entirely, but still grants medical experience? I would keep medicine as an influencing factor. Otherwise, absolutely love this idea.

moveup = (((doctor.mind.get_skill_level(/datum/skill/misc/medicine)) * 5) + moveup)
This is the part where it takes doctor's medicine skill to equation.

NGGJamie:
I was reading on mobile, I might've just missed that. LGTM.

PR #386, opened 2024-04-25
Swap chance of stomach crit with chest crit

panvxv:
## About The Pull Request
Making so hitting stomach has higher crit chance than hitting chest as stomach is harder to hit and crit to chest with a chop/stab kills.

## Why It's Good For The Game
Balance changes are good and healthy, especially with insta kill crits.


Ambrose997:
I feel this sorta ignores the fact in the code already has prevent_crits values if you use cut, chop and blunt, the only one it doesnt outright prevent crits from is stab, however it still has a 100 value for blocking melee damage, so rng comes a lot more into play when it comes to chest crits and stomach crits.

Personally I think logically a chest lethal stab is more likely to kill you than a stab in the stomach area in real terms.. if the code makes it less likely to hit the stomach area than the chest then I feel that's what needs to be changed, not crit chances.  

In regard to stomach crits having a lower chance, I assume this is because it quite literally removes your organs which makes healing even with a priest difficult/potentially impossible? From what I can see, it doesn't look like anastasis replaces organs.

Ambrose997:
Yes.. it'll kill you, but it wont make you unreviveable much like a gut spill would, but alas it's been merged.

PR #392, opened 2024-04-25
Adventurerslimitclasses

bobjension:
Really fucking weird there can be a infinite amount of paladins, clerics, rogues, mages, vaqueros, female mages, etc

however, the two classes that I think are really fucking basic are warrior and ranger, so they stay infinite

also adds 1 single slot to crusader, just for someone to dick around with him.

### DIFF
--- a/code/modules/jobs/job_types/roguetown/adventurer/types/combat/amazon.dm
+++ b/code/modules/jobs/job_types/roguetown/adventurer/types/combat/amazon.dm
@@ -5,7 +5,7 @@
 	allowed_races = list("Humen")
 	outfit = /datum/outfit/job/roguetown/adventurer/amazon
 	allowed_ages = list(AGE_ADULT, AGE_MIDDLEAGED)
-	pickprob = 13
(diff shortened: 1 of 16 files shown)


Wintermote:
I don't mind a limit (although I don't think it's necessary at all), but I don't want to select a character who's thing is being a cleric, only to find out there are no more cleric slots and now I have to play whatever else that doesn't fit that character at all.

If it's going to be like this, then the slots should be selected at round start rather than just being "adventurer" so I can select a different character to play a different (open) job instead.

BlueMemesauce:
this would make half the server warriors or rangers

bobjension:
I'll reopen it soon.

Ambrose997:
Imo really bizarre to limit them en masse like this. The main issue is having a potential for like 50 rogues or something retarded. 

Limit them, and limit others if the need arises (which is already a principle you guys usually follow).

CVSthePharmacy:
just forget this PR, community hates it

PR #394, opened 2024-04-26
nerfs bandits

emoats18:
gargle on mine knackers if thy do not like it!


CVSthePharmacy:
this guy's a cuck, might revert later

PR #403, opened 2024-04-26
Even more descriptions

Reyblood:
## About The Pull Request
- In this PR i add even more descriptions.
- I would like to point out that, in my last description PR, i described old turfs and it just so happens the game has a mix of the old and new ones, and so only the old turfs had description, now, i believe, almost all turf should have descriptions lest i missed something.
- For some reason, roguefloor.dm file gets indentation issues when i use the web editor. Will add descriptions to several turf objects later when i get access to a decent IDE, but the code should add descriptions to several objects and other turfs as well.

## Why It's Good For The Game
- Descriptions are good for the game.

### DIFF
--- a/code/game/objects/items/rogueitems/bags.dm
+++ b/code/game/objects/items/rogueitems/bags.dm
@@ -1,6 +1,6 @@
 /obj/item/storage/roguebag
 	name = "sack"
-	desc = ""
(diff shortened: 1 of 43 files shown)


ChungusGamer666:
Amazing PR, does breed life into the game. Any bad descs can be changed in a later PR, this is good as is.

PR #416, opened 2024-04-27
Fixes dezombification

ChungusGamer666:
At some point i forgot to remove the traits of zombies when the antag datum got removed, so every "cured" zombie was stuck in this permanent nightmare scenario of being a zombie but also not.

Removes TRAIT_LANGUAGE_BARRIER from zombies, because I think it's funnier if all their speech just gets replaced with "UUHHHHHHHHH..."


CVSthePharmacy:
once deconverted you have like, immortality to stomach slashes

PR #424, opened 2024-04-28
Stone Buffs

JTGSZ:
Increases stone throwing damage and regular damage.
Re-adds getting a random amount of stones from a rock
Re-adds gettin stones from things you are mining.

The numbers amount was supposed to come back forever ago cause it fucked masons over but we all forgot.


ChungusGamer666:
godly pr

Issue #292, opened 2024-04-21
Patron selection in the character creator reverts when round resets


Wintermote:
![Untitled](https://github.com/Blackstone-SS13/BLACKSTONE/assets/6611263/13ac54bc-71a2-44d7-bf06-ba518d316b44)

Selecting patrons different from Astrata will reset at the end of every round back to Astrata, even if you make sure you've saved your character with a different patron diety, (e.g. Necra).

Expected: Set Necra in character creator, get necra when loading that character
Actually: Set Necra in character creator, get Astrata when loading that character

This may be related to #275

Wintermote:
Seems to have been fixed by #432; closing this

PR #432, opened 2024-04-28
Fixing sharing common patron between your characters

AlexMorgan3817:
## About The Pull Request
If you try to switch character from list, you will have the same patron as your previous selected character.
Now it is fixed and I rewrite patronlist to have single patron's instance instead of creating patron for every players' character.

## Why It's Good For The Game
It saves memory, that was used for exact equal data storages called "PATRON", that was repeating for every character of same religion?
It let's player to not bother about reselecting patron every time he change character.

### DIFF
--- a/code/__HELPERS/global_lists.dm
+++ b/code/__HELPERS/global_lists.dm
@@ -71,9 +71,9 @@
 		GLOB.mouseicons_human += file("icons/effects/mousemice/swang/[i * 5].dmi")
 
 	// Patron Gods
-	for(var/gods in subtypesof(/datum/patrongods))
-		var/datum/patrongods/A = new gods()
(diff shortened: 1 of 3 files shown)


Wintermote:
Per #292 this was also happening when rounds restarted, which would by default set your patron diety to astrata regardless of what your character had saved before

AlexMorgan3817:
> Per #292 this was also happening when rounds restarted, which would by default set your patron diety to astrata regardless of what your character had saved before

I think I should test it with my PR

AlexMorgan3817:
> Per #292 this was also happening when rounds restarted, which would by default set your patron diety to astrata regardless of what your character had saved before

I can't recreate this issue with fix, maybe it is fixed

PR #434, opened 2024-04-28
Devotion progression fix

panvxv:
## About The Pull Request
Fixes the max devotion of combat classes so they can't learn as much as non combat classes.

## Why It's Good For The Game
We don't want to have Templars being equal to Priests.


ChungusGamer666:
Being handled by this PR - https://github.com/Blackstone-SS13/BLACKSTONE/pull/455

PR #437, opened 2024-04-28
Lets Paladins learn at least 1 spell and not have them nerfed to oblivion.

ariaworld:
## About The Pull Request
As title.

## Why It's Good For The Game
Good, there's someone who's main intent is now nerfing already nerfed things and nuking them to the void, so this is done to counter-act that, given they don't listen to anything head devs, devs and admins tell them.


Autisem:
if they didnt listen to devs could they not just deny the PR as it happens?

PR #443, opened 2024-04-29
Gives the merchant a shylock and another sales window

zakanater19:
## About The Pull Request

Makes the merchants area a little bit larger, gives him a shylock and also another sales window on the other side of the room

## Why It's Good For The Game

Long lines at the merchant, this should let shophands buy and sell on the other window to speed it up

Tested and working locally, also you can merge multiple map PRs and as long as they don't go over the same ground you can merge them all and it just works


zakanater19:
the reason you can merge multiple is because strongDMM automatically saves in TGM and it's compatible with the main branch server files, no idea why, just know that it works

PR #449, opened 2024-04-29
Shield Balance

Osmos123:
## About The Pull Request
Differentiates the shields stats though coverage changes (tower shield most coverage, balanced for defense while kite defense heavy). Shields no longer block 100% of everything when passing parry checks or for projectiles.

## Why It's Good For The Game
Makes the most basic shield no longer invulnerable to projectiles (theoretically, i am shit coder)

### DIFF
--- a/code/game/objects/items/rogueweapons/shields.dm
+++ b/code/game/objects/items/rogueweapons/shields.dm
@@ -29,7 +29,7 @@
 
 /obj/item/rogueweapon/shield/hit_reaction(mob/living/carbon/human/owner, atom/movable/hitby, attack_text = "the projectile", final_block_chance = 0, damage = 0, attack_type = MELEE_ATTACK)
 	SEND_SIGNAL(src, COMSIG_ITEM_HIT_REACT, args)
-	if(attack_type == THROWN_PROJECTILE_ATTACK || attack_type == PROJECTILE_ATTACK)
+	if(attack_type == THROWN_PROJECTILE_ATTACK || attack_type == PROJECTILE_ATTACK || prob(coverage))
 		if(owner.client?.chargedprog == 100 && owner.used_intent?.tranged)
 			owner.visible_message("<span class='danger'>[owner] blocks [hitby] with [src]!</span>")
 			return 1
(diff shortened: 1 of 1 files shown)


code/game/objects/items/rogueweapons/shields.dm
```diff
@@ -29,7 +29,7 @@
 
 /obj/item/rogueweapon/shield/hit_reaction(mob/living/carbon/human/owner, atom/movable/hitby, attack_text = "the projectile", final_block_chance = 0, damage = 0, attack_type = MELEE_ATTACK)
 	SEND_SIGNAL(src, COMSIG_ITEM_HIT_REACT, args)
-	if(attack_type == THROWN_PROJECTILE_ATTACK || attack_type == PROJECTILE_ATTACK)
+	if(attack_type == THROWN_PROJECTILE_ATTACK || attack_type == PROJECTILE_ATTACK || prob(coverage))
```
Autisem, line 32: would this not give a change to blocn any time of attack?

Osmos123: I believe this if statement handles projectile blocking, while the other if statement just below it handles other attacks given its just the coverage probabilty on line 37.

Autisem: ye but, since its OR it will ALWAYS pass the first or 2nd if if its a projectile and not even check the one your adding. This will only prob roll if its not projectile

Osmos123: ahhhh, makes sense, would this change work?

code/game/objects/items/rogueweapons/shields.dm
```diff
@@ -29,7 +29,7 @@
 
 /obj/item/rogueweapon/shield/hit_reaction(mob/living/carbon/human/owner, atom/movable/hitby, attack_text = "the projectile", final_block_chance = 0, damage = 0, attack_type = MELEE_ATTACK)
 	SEND_SIGNAL(src, COMSIG_ITEM_HIT_REACT, args)
-	if(attack_type == THROWN_PROJECTILE_ATTACK || attack_type == PROJECTILE_ATTACK)
+	if(attack_type == (THROWN_PROJECTILE_ATTACK && prob(coverage)) || (attack_type == PROJECTILE_ATTACK && prob(coverage)))
```
Autisem, line 32: either works tho this would be more readible, you must also test localy since that has become new policy
```suggestion
	if((attack_type == THROWN_PROJECTILE_ATTACK || attack_type == PROJECTILE_ATTACK)) && prob(coverage))
```

PR #453, opened 2024-04-29
Build Mode

NGGJamie:
## About The Pull Request

Re-enables build mode, with exception to Basic mode, as it still uses TG assets. Other modes also include TG assets, but can be used just fine with care. I'll keep this TMed so it can be used, but getting it to work fully will be a work in progress.

## Why It's Good For The Game

Better build capabilities for staff.


Autisem:
salute, maby good time to axe some TG turfs?

Real-MAGNUM:
Did some extensive usage live on the server during it's test merge (still active). Works great from what I've experienced.

PR #454, opened 2024-04-29
Repaths /datum/species/elf/snow to /datum/species/elf/wood

ChungusGamer666:
I don't like when the code lies to me. Snow elves are a separate thing entirely and might be actually implemented in the future.


code/modules/mob/living/carbon/human/species_types/roguetown/elf/elfs.dm
```diff
@@ -1,7 +1,7 @@
 /mob/living/carbon/human/species/elf/snow
```
BadAtThisGame302, line 1: You forgot this.

ChungusGamer666: OH TRUE

PR #460, opened 2024-04-29
Fixes naked and afraid paladins and templars

ChungusGamer666:
Seems like chimps here didn't know how associated lists work LOL


ChungusGamer666:
selfmerging idc

PR #466, opened 2024-04-29
Friendly zombie destruction

ChungusGamer666:
Friendly zombies are really starting to get on my damn nerves.

For every 2 minutes you don't bite a living human, one of your limb falls off - The last limb to fall off is your head, then you fucking DIE.


ChungusGamer666:
Does not apply to zombies without a client.

Alexeivichi:
Terrible way of doing honestly.

ChungusGamer666:
I have a better idea

OrbitalParadise:
Could you make it if a human enters a distance close to the zombie the timer starts, then the timer goes away if they are not close

PR #470, opened 2024-04-29
Fixes paper to actually utilise html formating (Episode 4560 of "Rogue Town code was a drunken mess")

Warfan1815:
## About The Pull Request

does what it says on the tin, now it uses the "write" hyperlink in the window instead of automatically writing (which causes issues)

## Why It's Good For The Game

Cool text


Warfan1815:
DO NOT MERGE I JUST REALISED I LEFT THE DEBUGGING MESSAGES IN

PR #473, opened 2024-04-30
Knives changes to be more in line with the other mechanics

panvxv:
## About The Pull Request
Made so their balance is "swift".
Weaponsmith has new recipy for "That's a knife" (https://www.youtube.com/watch?v=dSnosk4tWrg)
Removed chopping from elvish daggers and lower chopping cooldown a little bit.
Thrust now have more penetration than chopping.

## Why It's Good For The Game
So weapons aren't all the same, adding mechanics we have to current weapons makes combat more interesting.


CVSthePharmacy:
you mean stab?

panvxv:
> you mean stab?

yes

PR #475, opened 2024-04-30
Shield shall work correctly

panvxv:
## About The Pull Request
Make it so the block intent works, you can charge it and when it is charged range projectiles will be stopped.
Otherwise have a chance to block them, bigger shield means better chance.

## Why It's Good For The Game
I have gold fish memory but I remember this working in past years. Shields cannot be passive 100% blocker of everything.
Let's have players put some inputs in their gameplay.

### DIFF
--- a/code/game/objects/items/rogueweapons/shields.dm
+++ b/code/game/objects/items/rogueweapons/shields.dm
@@ -49,7 +49,7 @@
 	icon_state = "inblock"
 	tranged = 1 //we can't attack directly with this intent, but we can charge it
 	tshield = 1
-	chargetime = 0
+	chargetime = 1
 	warnie = "shieldwarn"
 
 /obj/item/rogueweapon/shield/wood
@@ -58,7 +58,7 @@
 	icon_state = "woodsh"
 	dropshrink = 0.8
 	wdefense = 15
-	coverage = 100
+	coverage = 40
 
 /obj/item/rogueweapon/shield/wood/attack_hand(mob/user)
 	if(!overlays.len)
@@ -96,7 +96,7 @@
 	wlength = WLENGTH_NORMAL
 	resistance_flags = FLAMMABLE
 	wdefense = 15
-	coverage = 100
+	coverage = 70
 	parrysound = list('sound/combat/parry/shield/towershield (1).ogg','sound/combat/parry/shield/towershield (2).ogg','sound/combat/parry/shield/towershield (3).ogg')
 	max_integrity = 200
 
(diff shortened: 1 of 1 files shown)


Gyrandola:
direction still needs to be added, you shouldn't be blocking a bolt coming sideways or from behind

panvxv:
> direction still needs to be added, you shouldn't be blocking a bolt coming sideways or from behind

will try working on this but first need this to be pushed. I think I will make so incoming projectiles from behind reduces the chance to block them.

Hmeister-real:
do shields on peoples back do anything to protect them?

Knightfall5:
Do we really want to nerf the only counter to ranged sniping? Bolts can 1 hit kill you through armor if they get a crit, roof snipers will get even worse if shields don't stop them.

YosemiteYam:
Only coverage of 40 for normal shields and 70 for the rest seem ass, shields are the only way to prevent you being one shot by a roofcamper right now. Before nerfing shields we should be nerfing the reasons they're actually used first. All this PR does is allow one shot crossbows to be able to now tap their only enemy, the 2% of people that have a shield and keep it out. I know you're gonna have a block "intent" work but honestly they should have a better passive defense and not have to rely on you charging up a block when you have it in one of your arms.

panvxv:
> Only coverage of 40 for normal shields and 70 for the rest seem ass, shields are the only way to prevent you being one shot by a roofcamper right now. Before nerfing shields we should be nerfing the reasons they're actually used first. All this PR does is allow one shot crossbows to be able to now tap their only enemy, the 2% of people that have a shield and keep it out. I know you're gonna have a block "intent" work but honestly they should have a better passive defense and not have to rely on you charging up a block when you have it in one of your arms. You're only not going to be able to use the block feature -in- combat while fighting due to using the weapon hand, but theoretically someone would be able to keep their shield up in a fight. Raise the passive coverage to something higher.
> 
> Also requiring to charge it inside of it just being on the active block input seems pretty imbalanced when shooting something is much easier, why is BLOCKING an arrow with a shield take so many extra steps versus actually shooting at it? You can look in the distance and shoot an arrow/bolt at someone that doesnt see you yet, if they dont have block actively charged that means they're gonna get shot lol.

70% chance to block a range a attack is ass?
And as how it works now, you just need to charge it once then you can just be on block intent without holding left click.

Knightfall5:
> 
> 70% chance to block a range a attack is ass? And as how it works now, you just need to charge it once then you can just be on block intent without holding left click.

What that actually means is you have a 30% chance to get hit with a bolt that will either kill you or give you a wound that is lethal if you don't retreat to heal. From someone you cannot even see because they're two z-levels up on a roof already reloading another bolt to shoot at you.

ChungusGamer666:
I want to make shields directional, this is overzealous

ChungusGamer666:
For now this is good. Shield improvements pending.

PR #476, opened 2024-04-30
Underworld Changes

NGGJamie:
## About The Pull Request

Adds more variety to spirit names so that when people ahelp "Dejected is [doing bad thing]" there won't be 20 spirits with that exact name. 100 different name combinations are possible.

Removes the ability for spirits to change movement speed or middle mouse intents.

Drastically reduces the number of tolls that need to be destroyed before triggering a respawn. Tolls will respawn when they hit 7 left.

## Why It's Good For The Game

Easier identification of specific spirits, easier for them to find tolls, should kill toll stealing except for real this time.


NGGJamie:
Should be good to merge whenever.

PR #483, opened 2024-04-30
Removing unused code and maps, also fixes map check

OzfurAtlas:
removed code we will not need;
space ruins,
old maps.

Fixes:
-fixed default map request for box station, now set to roguetown2


ChungusGamer666:
Testmerge seems very successful

PR #484, opened 2024-04-30
Enables grabber again, the lower merchants guild members

zakanater19:
## About The Pull Request

Adds four grabbers, lower members of the merchants guild, tested locally

## Why It's Good For The Game

Protection for the merchant

### DIFF
--- a/_maps/map_files/roguetown2/roguetown2.dmm
+++ b/_maps/map_files/roguetown2/roguetown2.dmm
@@ -3792,6 +3792,10 @@
 	},
 /turf/open/floor/rogue/cobble,
 /area/rogue/outdoors/town)
+"pb" = (
+/obj/effect/landmark/start/grabber,
+/turf/open/floor/rogue/wood,
+/area/rogue/indoors/town/shop)
 "pc" = (
 /obj/structure/stairs{
 	icon_state = "stairs";
@@ -75301,7 +75305,7 @@ sw
 sk
 sw
 sw
-sw
+pb
 sw
 oy
 sk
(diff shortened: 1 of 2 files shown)


NGGJamie:
What do these do that the existing two shophands aren't? We just had the PR propose 2 more and said that was too many.

zakanater19:
Well they can't value anything so they cant stand in the shop windows and trade, they can only be muscle and carry things up the balloon or do protection

TheManWithNoHands:
> They can only be muscle and carry things up the balloon or do protection

Those are things anyone can do. There is already the problem of people not interacting much with people outside their job bubble. This just makes it worse. Your making the extra muscels a merchant could buy (have RP interactions with other players) if he even needs it, into a role. 

A big reason why the shophand doesnt ad to this problem, is because the shophand assists the merchant but is also the learning role for that job. They have stuff to do, not any adventurer can do. 

I say, rather add a few spare keys this role would need, to allow the merchant to hire himself muscels more easily.

tl;dr Role bloat bad.

OrbitalParadise:
1 shop hand 1 grabber

YosemiteYam:
FOUR grabbers? That sounds terrible. Just let the merchant hire guards/mercs, they're the only people that got the resources to do so besides the king anyway instead of giving them shitty roles that can only be half-decent guards

NGGJamie:
As mentioned above, shophands basically already fill this role plus help at the windows. If the merchant needs a runner, they can send their shophand, or one of them, to perform the task. Additional roles would be overkill for that purpose.

PR #481, opened 2024-04-30
Class Archetypes -- Part One

TheVanderlin:
## About The Pull Request
Introduces a prompt attached to the Adventurer Class. After you choose a class -- certain adventurer classes are eligible for Archetypes. More will be added in future PR's. Archetypes are essentially variations of the original class, with the original also being selectable. Similar to traditional SS13 alt_titles. 

Deepened skills and added RNG. Certain classes and jobs were incredibly lacking in skills. Knights didn't know how to ride horses, the Queen didn't know how to read. Many combat classes lacked basic skill in many weapons, if you are a Sword master you should bloody well know how to use an axe or mace. 

Rebalance edits are mostly just to fix things that were out of whack. Trying to create a more centralized rule-system for how stats work. I generally kept to a 4/5 and a 7-8 stat system. Where the overall positive bonuses to stats were either 4-5 OR 7-8.  Whether they got the first range or second was down to what category of class it was, like a Steward being different to a Warrior. Cleric different to a Barbarian.

Fixed up traits, a lot of classes didn't know how to use medium armor despite having heavy armor proficiency. A lot of classes also lacked Steel_Hearted despite their role. Dungeoneer wasn't steel hearted? Whut.

Tried to make the Towner roles more fun, by giving them some more utility skills and some of them basic combat stuff. Cheesemaker I am most proud of, they have wonderful cheese constitution.

Mage and Court Wizard functionally aren't too different beyond having more skills and more interesting stat-design.

Cleric has two Domain Archetypes uniquely. War and Life Domains. If you choose War you are more or less a normal cleric, you can fight and beat people up. Life Domain gets extra utility spells and better clerical skills like Medicine. Their magic is also skilled higher.

Lastly out of just pure frustration at looking at Polearms in the game as I was adding them to Archetypes. I did some edits to make Polearms feel a little more unique. I really disliked how they were all essentially just Spears with a few extra intents. My intention later is to make a PR specifically to add more Weapon_Intents to the game to remedy this absolute horse shit.

## Why It's Good For The Game

Archetypes are the first stage in class/faction reworks in order to create variety in the game and introduce a more grounded style of RNG that isn't just a bunch of if(prob's) stacked on top of each other. 

The archetypes also allow for specialization of classes, instead of feeling like generic representatives of a fantasy trope. I also sought to deepen character classes and roles to feel more useful. By adding detail, and finding ways to make more classes and jobs feel valid options for players, we can make the game experience far better for the vast majority of players. 

[...]

PR #489, opened 2024-04-30
Staffed Monks

SELFHELL:
## About The Pull Request

Swaps out Monk's retarded CQC values for more interesting polearm ones.

## Why It's Good For The Game

Monks beating people to deaths with sticks is far more thematic than groin twisting and punching ones stunned body to death, also polearms are THE most used weapon of war throughout the ages, Ravox likes those.

### DIFF
--- a/code/modules/jobs/job_types/roguetown/adventurer/types/pilgrim/monk.dm
+++ b/code/modules/jobs/job_types/roguetown/adventurer/types/pilgrim/monk.dm
@@ -29,8 +29,9 @@
 	r_hand = /obj/item/rogueweapon/woodstaff
 	if(H.mind)
 		H.mind.adjust_skillrank(/datum/skill/misc/reading, 3, TRUE)
-		H.mind.adjust_skillrank(/datum/skill/combat/unarmed, 5, TRUE)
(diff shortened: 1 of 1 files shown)


Inanichts:
Monks are supposed to have unarmed skills actually, things like spells. Monks in D&D uses their fists like weapons and they can bypass armor with their punches and should be something close to that, i also agree to make it something that isn't just groin twist and punching them in the nose. Groins twist should be a torture thing where the player can't already do much. Monks can also twist other body parts to try and break arms/legs or the neck, but also they should be able to use staffs to hit people.

bobjension:
monks get strength and unarmed, no armor or anything else, this feels more like a cope PR

YosemiteYam:
Monks are supposed to only be using their fists...its the whole point of the class. They're pacifists who try not to kill you by only using unarmed techniques. Why give them polearms to kill people with?

Real-MAGNUM:
Please make a new adventurer class for polearms if you wish for staff-wielders. Our monks are D&D oriented. 

#481 might be worth waiting for due to protential conflicts, and it might let you do a sub-class (archetype) of monks. Everyone would be happy that way, yeah? 

I like your idea, but I think it's best to add a subclass to monk in this situation, instead of replace. Closing PR.

PR #491, opened 2024-05-01
Revert baliffnerd

bobjension:
snowflake role and his goons GTFO

PR #493, opened 2024-05-01
nerfs bailiff and councilor 

bobjension:
since I know a CERTAIN SOMEONE will never revert bailiff and councilors, I will seek to make them not as godtier anymore.

bailiff got 40 damage mace roundstart? with 5 strength and 1 fortune, even sheriff isn't this good.
councilors got 2 fortune and monk unarmed stats with insane athletics
headdev
this is your headdev btw.

### DIFF
--- a/code/modules/jobs/job_types/roguetown/nobility/bailiff.dm
+++ b/code/modules/jobs/job_types/roguetown/nobility/bailiff.dm
@@ -10,6 +10,7 @@
 	"Half-Elf",
 	"Aasimar")
 	allowed_sexes = list(MALE, FEMALE)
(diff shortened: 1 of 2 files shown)


YosemiteYam:
Good pr, should honestly just get rid of the bloat roles that don't contribute anything but take point from all the other court roles away. Could you also revert the sheriff description to the old one instead of the shit they have right now that only talks about the bailiff, and make the bailiff description not be the longest fucking description by like an extra four sentences?

Real-MAGNUM:
Looks good to me. 

#491 Was closed due to bad PR format, and essentially no explanation as to why, not because it's forbidden to revert this addition. Discuss removing councilor altogether in the discord, it's being considered substantially more than bailiff, but we're open to reverting both.

PR #500, opened 2024-05-01
Needed map tweaks

Real-MAGNUM:
## About The Pull Request

Removes spawn points inside of the lava cave.
Fixes lava cave travel tiles.
Fixes missing turfs in the forest
Fixes the upper floor of the guard post at the bog gate by adding a stair

## Why It's Good For The Game

Bugs that I havent fixed yet but are getting annoying to fix with admin commands.


Real-MAGNUM:
Speedmerging because it's basically only a cumulative 20 tile bugfix.

PR #503, opened 2024-05-01
Balancing the hand of heartfelt.

Alibouk1:
This ought to make hand of heartfelt useful for the lord of heartfelt as opposed to simply a victim.


Real-MAGNUM (approved):
Finally a rebalance PR that isnt giving a class a +5 stat or skill. 

Looks good to me.

kacperpl1234pl:
very good pr i spawned once as hand of heartfelt and had 6 str and i was unable to use my sworld very traumatic exprience

PR #506, opened 2024-05-01
Gold Rings Price Nerf + Craftable Rings (Gold/Steel)

Sarkness:
## About The Pull Request

This pull request is after many complaints about the gold rings making the Economy crashes after players found you you can sell gold, buy gold and repeat the process (mainly with Steward, a smith and merchant)
This also adds the Emerald Ring that may be craftable if players are lucky enough to find an emerald gem (named Gemerald) and a gold ingot

## Why It's Good For The Game

Makes economy somewhat more stable + Adds the possibility to have Portable SCOMs more open to everyone (besides having it only for King and Heartfelt)


OrbitalParadise:
I think emerald rings should be considered a magic item, the lack of communication can be solved in the future by use of magic or an artificer. If everyone has emerald rings it's going to shit up everything

NGGJamie:
Agree with the poster above, it would be too much if everyone was running around with SCOM rings. I would be completely down for having craftable emerald rings that don't have the SCOM abilities, though.

Sarkness:
Fixed, Rings are now non-SCOM Rings

code/modules/clothing/rogueclothes/rings.dm
```diff
@@ -20,7 +20,7 @@
 /obj/item/clothing/ring/gold
 	name = "gold ring"
 	icon_state = "ring_g"
-	sellprice = 70
+	sellprice = 47
```
BadAtThisGame302, line 23: Should be 50 or 45, very weird to leave it on a 7.

Sarkness: I mean, it's more to round it to 141

UbaserB: > I mean, it's more to round it to 141

that is also equally weird

code/modules/roguetown/roguejobs/blacksmith/anvil_recipes/valuables.dm
```diff
@@ -45,3 +45,46 @@
...
+
+/datum/anvil_recipe/valuables/topazg
+	name = "Gold Topaz Ring (+ Topaz)"
+	req_bar = /obj/item/ingot/gold
+	additional_items = list(/obj/item/roguegem/yellow)	
+	created_item = /obj/item/clothing/ring/topaz
+
+/datum/anvil_recipe/valuables/emerings
+	name = "Silver Emerald Ring (+ Gemerald)"
+	req_bar = /obj/item/ingot/steel
```
JTGSZ, line 69: steel ingots for creation of silver rings, unless these are supposed to be steel rings in which case you change the name.

Sarkness:
Unfortunately there's no silvers yet, so if needed you can just remove it I
guess.
Otherwise until silver is really added I suppose?

PR #508, opened 2024-05-01
Update necromancer.dm

Ambrose997:
## About The Pull Request

Lets Aasimar play necromancer.

## Why It's Good For The Game

Couldn't find any particular reason why this class was restricted from the role. Maybe due to the weird half-angel lore? But on that same note Aasimar can be rogues, assassins, zombies, vampires, why would necromancer be any different? It's also not unbalanced statwise cus Aasimar get shit constitution and endurance, both which is penalized for necromancers even further, and Aasimar have INT comparable to Dark Elves so it surely isn't that.

### DIFF
--- a/code/modules/jobs/job_types/roguetown/adventurer/types/combat/rare/necromancer.dm
+++ b/code/modules/jobs/job_types/roguetown/adventurer/types/combat/rare/necromancer.dm
@@ -5,7 +5,8 @@
 	allowed_races = list("Half-Elf",
 	 "Dark Elf",
 	 "Tiefling",
-	 "Humen")
+	 "Humen",
+     "Aasimar")
 	outfit = /datum/outfit/job/roguetown/adventurer/necromancer
 	pickprob = 30
 


SELFHELL:
The "angel" lore was why I didnt allow it for them. The pantheon doesn't like "angels" actively pissing off necra/the cycle of souls or something. Also the edgy fallen angel motif is lame. 

Indifferent besides that

Ambrose997:
> The "angel" lore was why I didnt allow it for them. The pantheon doesn't like "angels" actively pissing off necra/the cycle of souls or something. Also the edgy fallen angel motif is lame.
> 
> Indifferent besides that

I hear ya, but again they can still be vampires, they can still be thieves/assassins/bastards, so im not really sure what the point is.

Ambrose997:
> The "angel" lore was why I didnt allow it for them. The pantheon doesn't like "angels" actively pissing off necra/the cycle of souls or something. Also the edgy fallen angel motif is lame.
> 
> Indifferent besides that

on that note the lore still seems to want them to behave like humans, since they were raised by em.

PR #511, opened 2024-05-01
Adds some blacksmithing gear in the sewers, also removes the impassible rock behind the prison cells and replaces it with dirt

zakanater19:
## About The Pull Request

Adds a few blacksmithing things in the sewers, you still need to bring your own blacksmithing TOOLS but the equipment is down there, also replaces the impassible rock behind the prison cells with dirt, so you can break the walls there and rescue a prisoner from the garrison prison, without touching the castle prison

## Why It's Good For The Game

Gives people a way to rescue prisoners that does not involve storming their way in through the front

### DIFF
--- a/_maps/map_files/roguetown2/roguetown2.dmm
+++ b/_maps/map_files/roguetown2/roguetown2.dmm
@@ -2245,6 +2245,10 @@
 "iD" = (
 /turf/open/floor/rogue/cobble,
 /area/rogue/outdoors/rtfield)
+"iE" = (
+/obj/item/reagent_containers/glass/bucket/wooden,
+/turf/open/floor/rogue/cobble/mossy,
+/area/rogue/under/town/sewer)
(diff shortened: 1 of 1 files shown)


YosemiteYam:
Why would the town dungeons, one of theoretically the most protected/hard to access areas, be easier to break into than a normal building with wooden/stone walls lmao? I can't see this ever actually contributing to an in depth roleplay situation of prisoner break out, instead it just lets people break into the town watch in two seconds if you have a pick or just grief the cells to be un-usable.

Real-MAGNUM:
Not very into this at all. 
We dont need another blacksmith, there will be one in the bog map.
Storming the front is more theatrical for everyone, and makes the most sense. I'll be replacing the indestructible rocks on the entire roguetown map with a destructible version eventually, however.
-1. Suggesting a close.

PR #518, opened 2024-05-02
Add Clerk Role - Learning Role for Steward

NGGJamie:
## About The Pull Request

Steward seems to be the last of the unique roles without an equivalent learning role. This adds a new 1 slot youngster role so they can give it a try.

Clerks do not get a steward key, and need to borrow from the steward if they wish to use the nerve master.

## Why It's Good For The Game

Learning is good.


Real-MAGNUM:
I'm in support of this. 👍

PR #523, opened 2024-05-02
Feint rework

panvxv:
## About The Pull Request
Feint can be always done with right click when combat mode is on.
Feint now has formula for success chance when we are more skilled than opponent.
Defend intent will be the most basic intent without downsides and giving you no cooldown on parries and dodges.

Defend will be useful most of the time when you are against someone with a fast weapon as most weapons don't attack fast enough to strike you when you are on cooldown. Defend will be also useful against more than one person.

## Why It's Good For The Game
Making feint actually be useful.

### DIFF
--- a/code/game/objects/items/rogueweapons/rmb_intents.dm
+++ b/code/game/objects/items/rogueweapons/rmb_intents.dm
@@ -55,7 +55,10 @@
 				if(I?.associated_skill)
 					theirskill = L.mind.get_skill_level(I.associated_skill)
 		if(ourskill > theirskill)
-			perc += ((ourskill - theirskill)*10)
+			if(istype(user.rmb_intent, /datum/rmb_intent/feint))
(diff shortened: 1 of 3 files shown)


zakanater19:
changing the meta

code/game/objects/items/rogueweapons/rmb_intents.dm
```diff
@@ -55,11 +55,9 @@
 				if(I?.associated_skill)
 					theirskill = L.mind.get_skill_level(I.associated_skill)
 		if(ourskill > theirskill)
-			perc += 100
```
ChungusGamer666, line 58: holy shit this was rough

ChungusGamer666 (changes requested):

CVSthePharmacy:
this fucks up the entire purpose of panx's knife pr, which was to let you stab through parries since it was so fast

PR #524, opened 2024-05-02
[/tg/station sends its regards] Make client rendering less ass (Re-adds client fps game preference, sets default to 100fps)

MrStonedOne:
Not sure what the hell is going on here but client fps isn't known to impact the server load at all so there isn't any reason to disable and gimp it like it was.

100 was chosen because 100 fps renders better than 60fps or 120fps in byond because of rounding with milliseconds that puts 60fps out of lockstep with the server's 20fps that doesn't happen when client fps evenly multiples with both 1000ms and the server's 20fps. This matches the default everywhere else.

from /tg/cord: 
image

This PR is untested.

### DIFF
--- a/code/modules/client/preferences.dm
+++ b/code/modules/client/preferences.dm
@@ -108,7 +108,7 @@ GLOBAL_LIST_EMPTY(chosen_names)
 
 	var/list/ignoring = list()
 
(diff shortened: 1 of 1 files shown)


emoats18:
they did this to hide that it was tg code so I've heard

PR #538, opened 2024-05-02
Cleric Adjustments

Osmos123:
## About The Pull Request
There is nothing holy about dark elves in the lore. They should not be allowed to be clerics.
image

## Why It's Good For The Game
Lore+Balance (Dark elf females are the only race that gets a +4 net gain to stats compared to other races.)


Real-MAGNUM:
Dark elves can barely play anything as it is. Lets give this one some thought first.

Knightfall5:
If a specific race is getting too many stats the solution is to bring that more in line with the other races instead of arbitrarily cutting them from playing classes

ChungusGamer666:
I get the logic here, but I am not a huge fan. Clerics do not have to be good people, nor worship good gods.
Consider locking delfs to "bad gods" like Xylix and Dendor instead? Idk

PR #544, opened 2024-05-03
Vault Economy Tweaks

NGGJamie:
## About The Pull Request

If an item worth 0 was placed into the vault, the vault would stop processing early and not pay out.

Adds a penalty for multiple of the same item type in the treasury, encouraging stewards to diversify the items they place inside. Each item of the same type will earn 33% cumulative less worth per duplicate, with the first earning full value, then each subsequent losing 33%.

Also, Stewards and Clerks will be alerted to the treasury accruing value.

Coins in the treasury will not accrue interest.

## Why It's Good For The Game

Encourages Stewards to diversify investments and not spam one item into the vault. Doing so will lead to severe drops in value.


NGGJamie:
Appears to be working well and not blowing up the economy.

PR #545, opened 2024-05-03
[FIX] Remove "suspiciouspicious" 

Gyrandola:
## About The Pull Request

Removes the "sus" and "sussy" accent autoreplacer.

## Why It's Good For The Game

Nobody says "sus" in 2024, yet a lot of people say suspicious or otherwise use words with "sus" in it. This anti-sus measure hurts a lot more than it helps as it prevents you from saying a lot of words.


NGGJamie:
this is sus but I'll merge it

PR #552, opened 2024-05-03
Staff and polearms make you faster on rough terrain but no longer polearms on backs

panvxv:
## About The Pull Request
As in title, polearms are really powerful weapons and for such reason and the fact they are longer than user, people no longer will be able to put them on their back.
BUT! Now walking will be all the same no matter the terrain when we have them in one hand.

## Why It's Good For The Game
Balance! Balance! BOG map not being as annoying!

### DIFF
--- a/code/game/objects/items/rogueweapons/melee/polearms.dm
+++ b/code/game/objects/items/rogueweapons/melee/polearms.dm
@@ -108,7 +108,6 @@
 	gripsprite = TRUE
 	wlength = WLENGTH_GREAT
 	w_class = WEIGHT_CLASS_BULKY
-	slot_flags = ITEM_SLOT_BACK
 	minstr = 8
 	max_blade_int = 100
 	anvilrepair = /datum/skill/craft/weaponsmithing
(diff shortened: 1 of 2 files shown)


TheManWithNoHands:
I.. i dont really get why holding a pole arm would make your speed the same no matter the terrain.
Mind explaining that a teny tiny bit?

I would get if you could charge with polearms, but just gettig a dam speed buff because your holding a large weapon that is too large to carry on your back seems.... strange.

ChungusGamer666:
Not a fan

Wintermote:
This was a real issue with them IRL, you had to carry the damn things everywhere, same was true for greatswords/zweihanders/montante/spadone. It's one of the big reasons you didn't see them outside battlefields and everyone opted for sabers, rapiers, smallswords, etc.

Greatswords in particular were largely used in pike formations and by bodyguards; in pike formations they had an advantage over spears in weight balance and leverage and featured parrying hooks and a crossguard both of which helped bind, control, and parry spears in contact with the blade, which enabled greatsword users to open pike formations for friendly pikemen to exploit.

In the context of bodyguards, the reach of a greatsword and it's weight and momentum, made it possible for one or two wielders to hold multiple fighters back - with the weight and momentum behind the blade (it was used in constant sweeping/dancing motions) making it very difficult for smaller and lighter swords to parry.

This is a video that (despite being re-enactment and not sparring) shows generally how greatswords were used: https://youtu.be/vxHaNRO705k?si=K2cGyqfLa0Ko_8ov

ChungusGamer666:
the pr that added polearms on back was sneaky about it and labelled it a QoL, that shit is downright satanic and this PR will be merged to rectify that sin

PR #554, opened 2024-05-03
Knights in full suits of armor shouldnt be able to climb walls

OrbitalParadise:
## About The Pull Request

title says it all, discuss

## Why It's Good For The Game

sir sexton nerf


YosemiteYam:
This is braindead, you're nerfing the guy inside the armor to not be able to climb, not dudes in armor. Everyone else would still be able to climb if they're in plate but not a knight. 

Also the knight role is literally supposed to be the super role when it comes to athletics, ofcourse they can climb. What possessed you to make this pr, what sort of machiavellian plot will a discussion about this lead to

OrbitalParadise:
This PR is under the assumption that knights will always be in armor which is usually the case, I know the rest of the skills like athletics would fall into question. However if they really are a super role why the -2 to speed, cause the devs made it so they assumed knights will be in armor always.

OrbitalParadise:
It would be better to adjust skills based on wearing armor but that wont be for like a long time

ChungusGamer666:
No, but a heavy armor training check would be good

PR #557, opened 2024-05-04
waterskins

cornercube:
## About The Pull Request
adds a leather waterskin
holds 64 units (game lists it at 21oz), twice a mug, a bit less thana bucket
classed as bulky, does not fit in your backpack but it will sling around neck or belt 
future update will add the flask (when sprite is available) for water source that fits in your bag

crafting recipe: 3 hides, 2 threads

barp

credit to mushroomfrog2779 for sprite

theres also a filled and unfilled sprite added, but only the filled sprite is used for now
no idea how to code sprite update

## Why It's Good For The Game
you ever tried walking a kilometer with a mug in your hand trying not to let it spill?


Wintermote:
There are actually bottles already that you can place in bags and carry water in (15 units)

YosemiteYam:
Could you add it as an item for the merchant shop to order too?

> There are actually bottles already that you can place in bags and carry water in (15oz)

Yeah but you can't make bottles can ya? They're a bit rare I feel. (Maybe we can add a glassblower as a profession to make glass items)

cornercube:
> Could you add it as an item for the merchant shop to order too?
bottles, sure
was planning on making them craftable, but we do not actually have any silicon-bearing goods on the map like sand

for the waterskin itself, maybe leather

Real-MAGNUM:
Please make this craftable with something. Leather would work fine if we have that.

PR #559, opened 2024-05-04
Fixes mercs starting with no lit torches in the bog, also fixes gretzlemercs starting with no polearm due to a bug

Alibouk1:
The bug is due to making Polearms not fit onto the back, which means that the attempt to spawn them on the back fails, I moved them to the right hand, torches are added due to them spawning in the bog, I know their mercenary base starts with torches but it could be too dark to see, or they could be stolen by others, so I feel just starting with them lit is way more effective and intuitive.


Androidonator:
doesnt work if its merged and deployed it doesnt work

Alibouk1:
I'll look into it again.

Alibouk1:
dreamseeker_9HmRBZuLzm
Tested it again, should be working just fine. @Androidonator

PR #560, opened 2024-05-04
Nudists can't wear armor, necrophiles are zombie infection immune, deslops trait code for advclass

ChungusGamer666:
slop slop slop slop


dunsel-hordon:
your change fucked it all i can't put on anything as nudist this needs to be reverted

PR #563, opened 2024-05-04
enables russian accent for wood elves

ChungusGamer666:
funny, also very minor as i checked the json file


Real-MAGNUM:
Testmerging.

PR #565, opened 2024-05-04
Adds arrows recipe

Alibouk1:
Bolts are craftable by the smith, but iron arrows are not, I don't see a reason for this.


Real-MAGNUM:
Testmerging.

PR #570, opened 2024-05-04
Zombie re-adjustment number 5001

JTGSZ:
Zombies lose the trait NOLIMBDISABLE.


ChungusGamer666:
yeah baby!!!!

dunsel-hordon:
while i get wanting to nerf zombies a bit this adjustment really isn't the way to go. You are a zombie you don't feel things wrong with you. I could understand if a limb got crit broken then sure but being in pain and the limb not working makes no sense.

PR #575, opened 2024-05-05
Return of:  no skill in smithing = bad repair

panvxv:
## About The Pull Request
As it was at the start. People without skill shouldn't be able to repair their gear.
A change to how much one hit repairs as it was too much. Skilled will repair 15% with every hit, Legendary will repair 30%.

## Why It's Good For The Game
So skill matters and you need others to help you.


CVSthePharmacy:
so can you no longer learn smithing

PR #586, opened 2024-05-05
Triumph buys framework, Pilgrim and Adventurer Merger, Cleanup, Menus, (Playable Goblins)

JTGSZ:
### **-DO NOT TESTMERGE-** this unless you want to watch the subsystems/profiler yourself.

This PR does the following, 
Merges pilgrim and adventurer into one job. Vagrant, who will get a few classes from both on the list.
Adds a browser popup menu for class selection and attempts to resolve the stupid azz out of slots classes being on the list.
Condenses the bools into bitflags for categorization on the class datums.
Retypes beggars from vagrants to beggars on their path as the new job is going to use the word. 
Dwarf Vagrant is back to just being a Dwarf Dwarf
Makes a Role Class SS + Expands the triumphs SS
Adds a triumph buy menu which comes with four testing examples.
(Grenzelhoft Only for round effects, playable goblins/pick any class for char effects, and wipe all triumphs)

Basically its mostly framework/cleanup, 
This is just a slopped out skeleton of things. It definitely is not the best/will probably have to be rewritten later anyways.

Also goblins bro....
Untitled


Wintermote:
I don't want to see characters who are typically playing the priest, now being a beggar or a townsguard depending on whether they win or lose their RNG roll (albeit this only affects adventurer right now?); it entirely circumvents the entire purpose of the job priority system that already exists. It diminishes roleplay, and gamifies the job and character selection like it's an arcadey roguelike rather than a roleplaying game.

Not everyone wants to be a random-name playing whatever job they can get; people want to create consistent backstories for RP with particular characters and this PR is the antithesis to that.  I understand that you want to reduce the number of people playing as cool warriors and stuff, but this, as implemented, isn't the best way to do that. 

Frankly I'd have just taken individual classes out of adventurer and made them separate jobs, the same way you roll separately for merc and bog guard. This lets you control how many people are being whatever class and provides them the opportunity to return to lobby and play a different job/character when they lose the RNG roll.

The triumph shop, I think is a good addition, but shouldn't be related to 'purchasing' job roles.

At a bare minimum I'd like an option to choose a different character if I don't get the roll for the job I want to play a character with. I don't want to roll for warrior with my warrior character and get cheesemaker, it just doesn't make sense to play a consistent character with inconsistent random jobs.

pismorizzo:
> It just doesn't make sense to play a consistent character with inconsistent random jobs.
> 
> Triumph shop good
> 
> but if I lose the RNG roll for a job I want to return to lobby to play a different character, which is what the existing job selection roll already accomplished.
> 
> IMO just split adventurer classes into separate rolls the way mercs and bog guards are, if you think there's too many people being special snowflakes, it would work better than this.

This exactly when I roll adventurer I already just want 1 specific role adding another 12 awful bloat rolls to it just means I cant play what I want and am forced to kill myself to get another shot at it.

Yellows91:
Suggestion. (Bastard/disgraced son triumph spawn)

TheManWithNoHands:
Hey, question. You said this is a framework/cleanup. The adventurer classes are MUCH stronger then pilgrems stat wise. Being that pilgrems are more crafters while Adventures are Battle classes. With the slots for adventures capped to not flood the town with battle casts and fuck up any balance.

So how can the merge of both help balance wise? I kinda dont see how if you tie the infinit latejoin role to a capped battle role. But i migth have just missunderstood something, so.

Real-MAGNUM:
Vagrant job needs to be uncapped.
Let people spend 1 triumph to reroll or 3 to straight up pick their class. Something like that idk.
No merge till triumphs are used to reroll/select class imo. If it's too cheap, we'll increase the cost.

JTGSZ:
Closing this for now so someone don't do a funny

PR #589, opened 2024-05-05
Seizing the means of adminwho() (Available to players now)

Ambrose997:
## About The Pull Request

Allows us to see which admins are ignoring ahelps, shitters beware :)

## Why It's Good For The Game

Self-explanatory. The restriction is random af and I don't see a justifiable reason for it to exist. We should all be able to see which admins are online. At least now I know how to call out for ignoring ahelps if they're smart enough to avoid announcing their presence.

### DIFF
--- a/code/modules/client/verbs/who.dm
+++ b/code/modules/client/verbs/who.dm
@@ -99,10 +99,7 @@
 	set category = "Admin"
 	set name = "Adminwho"
 	set desc = "Lists all admins currently online."
-	
-	if(!check_rights(R_ADMIN))
-		to_chat(src, "You do not have the rights to use this command.")
-		return
+
 	var/msg = "<b>Current Admins:</b>\n"
 	if(holder)
 		for(var/client/C in GLOB.admins)


BlueMemesauce:
We should also steal the thing where it pings admins on discord if there are none online

NGGJamie:
The reason it's restricted is primarily to stop rule breakers from knowing exactly how many staff are online to coordinate when they're going to do their thing. Secondarily, there's not a good reason for players to know which staff members to put down, because all they know is that staff member is online, but they don't know anything else.

Maybe they're already answering a ticket and that ticket just requires a ton of investigation. Maybe they're running an event that requires a lot of attention. Maybe they dedicated the last round or two to answering tickets and they just want a round to play an important role without stepping away. Maybe they're like me and keep the game on in the background while they do development and just check up occasionally.

Plus, it's not even as if tickets frequently go unanswered. When there are, it's usually because tickets flood in, only 1-2 staff are online, and by the time you get into them, the later ones are so old that nothing can really be done for them anymore. If we do have staff that aren't pulling their weight in tickets, then we'll address them as it comes up. I can say from experience outside Blackstone that other staff certainly notice when their peers are consistently not doing tickets, and they don't like being the one carrying.

I enabled this verb so admins could have a better idea of when they're alone and give them the ability to call for help. There's really not a significant benefit to extending the verb outside of staff.

Real-MAGNUM:
Adminwho is remaining private. Jamie did a very good job explaining why.

PR #599, opened 2024-05-06
unhardcodes lobby music

JTGSZ:
does exactly whats on the label, moves the old track to the folder and renames it to title1.ogg and also lets music play from the folder


Real-MAGNUM:
Testmerging

PR #610, opened 2024-05-06
Mining ease

Osmos123:
## About The Pull Request
Cuts the rock wall health in half in consideration to the strength nerf (30%>10%). Hopefully encourages people to mine and deposit the ore now.

## Why It's Good For The Game
Very annoying to mine.


panvxv:
Currently you need 5 hits to destroy a rock with 10STR. Why would you make it 3 hits?
With 12STR it only takes 4 hits.

PR #611, opened 2024-05-06
Gives Princess shortboots and the Prince noble boots and equalizes the Prince/Princess mammon 

BadAtThisGame302:
## About The Pull Request

Gives the Princess shortboots so she doesn't walk around barefoot

Gives the Prince the same amount of mammon the Princess gets

Gives the Prince noble boots

Removes an extra dot from the description

## Why It's Good For The Game

Princess should not walk barefoot

Prince not having any mammon and the Princess having a pouch full is not very equal

The Prince is a noble, obvious choice for boots

Extra dot, fixed for better grammar.


JTGSZ:
I think the princess being barefoot was deliberate on the part of the predecessors, but It doesn't really matter at all here.

PR #612, opened 2024-05-06
Made Queen her own code and made her a noble

BadAtThisGame302:
## About The Pull Request

Gave the Queen her own flag instead of being a serf merchant

Also gave her the noble trait, since the description states she was picked out for political value

(Removes an extra dot in the desc)

## Why It's Good For The Game

Queen should both have her own code and not leech off others 

Plus, she should be a noble like her husband and kids

(Grammar)

### DIFF
--- a/code/__DEFINES/jobs.dm
+++ b/code/__DEFINES/jobs.dm
@@ -96,12 +96,13 @@
 #define NOBLEMEN		(1<<0)
 
 #define LORD		(1<<0)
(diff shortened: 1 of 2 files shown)


JTGSZ:
Well, she is marked as nobility in the pathing, and also has ruler code. So correct interpretation. However you removed a commented out useless hat, so I am going to go ahead and commit it back into ur pr so she gets it if she doesn't come out as the ruler of the session.

PR #613, opened 2024-05-06
Inconsistency between Weapon and Armor smithing steel numbers

wgtjunior743:
## About The Pull Request

This is a basic change to the numbers fixing what weapons had mostly right but armor was wrong, as a blacksmith you are already using a steel to see that item, so it makes sense to show only +steel if you are using only one extra or +2 steel if you are using two extra.

## Why It's Good For The Game

I experencied myself as blacksmith that the difference between numbers was quite confusing, even more for a first timer, sometimes overcharging clients because i thought you would need to use more steel than the number truly was.


JTGSZ:
I see the logic in this.

PR #618, opened 2024-05-06
Knight Gatemaster keyGODS, Bailiff Counciller keyNERDS

bobjension:
adds knight keyring, (they're never getting personal rooms)

also renames the "bailiff" to the real bailiff key.

removes wall key from most keyrings, key is never used and it should be reserved for the gatekeeper

also removes keys from bailiff and councillers because why would booknerds that aren't the steward and the archivist get access to the garrison, the manor AND the dungeon.

bailiff and councillers also don't live in the graveyard.

### DIFF
--- a/code/game/objects/items/rogueitems/keyrings.dm
+++ b/code/game/objects/items/rogueitems/keyrings.dm
@@ -2,7 +2,7 @@
 
 /obj/item/keyring
 	name = "keyring"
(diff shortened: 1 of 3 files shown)


bobjension:
oh wait they spawn in the manor, why did nobody bother making a place outside of town for them, I'll reopen this

PR #627, opened 2024-05-07
Magic Scrolls

Real-MAGNUM:
## About The Pull Request

Adds framework for magic scrolls and learnable spells via magic scrolls.

## Why It's Good For The Game

Spells cant be acquired outside of spawning with them.

Thanks to @solart3ch for helping!


Real-MAGNUM:
Tested in-game. Works fine. Too sleepy to do a recording right now.

JTGSZ:
Could use a sliding scale of int required for it depending on spell EX: fireball to greater fireball, but what is here is better than nothing in the first place.

PR #590, opened 2024-05-05
Racism lore: (Almost) Every nobility position is humen-locked, (almost) every other job has no race locks

ChungusGamer666:
## About The Pull Request

Rockhill has fallen...
image

What the title says. Every noble position is now humen only, EXCEPT for...
-Prince who can also be half elf because bastard children are hilarious
-Court mage because elves are magic and this is a classic fantasy trope
-Merchant, who can also be a half elf because again bastard children are hilarious

## Why It's Good For The Game

Better racism experience

PR #630, opened 2024-05-07
Re-add elf queens, remove dark-elves allowed in everything, and removes female squires

BlueMemesauce:
## About The Pull Request

Elf queens are very based and the description for queen says they were married for political reasons. 

Reverted #590 adding dark elfs and tieflings to everything, the whole point is that they're shunned in society, you literally get negative moodlet for looking at them. So they wouldn't be allowed as butler. And prisoner doesn't make sense for them either since they would just be executed and wouldn't be kept as a political prisoner. They definitely wouldn't be allowed to be the bogmaster.  I left them for blacksmith which is kind of borderline since blacksmith is a respected position but it's still a peasant position.

Removes female squires since the whole point of squires is that they are training to be a knight, which can only be male. 

## Why It's Good For The Game

How can you not like elf queens they're just very based.

### DIFF
--- a/code/modules/jobs/job_types/roguetown/garrison/bogmaster.dm
+++ b/code/modules/jobs/job_types/roguetown/garrison/bogmaster.dm
@@ -12,8 +12,6 @@
 		"Dwarf",
 		"Aasimar",
 		"Half-Elf",
-		"Tiefling",
-		"Dark Elf",
 	)
 	tutorial = "You are the most experienced idiot to volunteer to the Bog Guard... What a mistake that was. You report to the Sheriff, and your job is to keep the bogmen in line and to ensure the routes to the keep are safe. May the ten have mercy on you..."
 	allowed_ages = list(AGE_ADULT, AGE_MIDDLEAGED, AGE_OLD)
--- a/code/modules/jobs/job_types/roguetown/nobility/lady.dm
+++ b/code/modules/jobs/job_types/roguetown/nobility/lady.dm
@@ -7,7 +7,9 @@
 	spawn_positions = 1
 
 	allowed_sexes = list(FEMALE)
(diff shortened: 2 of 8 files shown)


ChungusGamer666:
elf queens are bad imo sorry

Gyrandola:
`And prisoner doesn't make sense for them either since they would just be executed and wouldn't be kept as a political prisoner. `

Disagree. First of all not all prisoners are political lol. Second of all "they would just be executed" is your headcanon.

JTGSZ:
Once again, conflicts and you need to talk to mister chungus about this.

BadAtThisGame302:
The only two bad changes is the Elf Queen and removing the Tiefling/Drow from Prisoners, it's weird race locking a role who is supposed to be stuck in the dungeons the entire round to only some specific races, like, let him be who they want.

Second, elf Queen is bad and I don't see how marrying an elf for "political" reasons is a good idea, might even be political suicide in a place like this, considering the only Elvish Kingdom in lore is not even close to the Isle and it would be very much worth it to marry a Humen Noble from either of the empires.

BlueMemesauce:
Keeping a prisoner was expensive since you had to feed them at least a minimum amount and keep them dying and give them free shelter. Dark elves and tieflings just aren't valuable enough to keep in a prison, they would probably be executed if they were suspected. Prisoners that were kept long term were usually important nobility that were being kept as hostages 

Foreign marriage creates alliances, which is pretty important for a weak kingdom. And elves and humans can create children. But I guess it wouldn't make sense if the elf kingdom was too far away

Real-MAGNUM:
Yes and no. So many changes I'd rather just make my own pr than spam this one with change requests.

BlueMemesauce:
when no elf queens
yakuza

PR #631, opened 2024-05-07
Traits barbarian.dm

GetProactive:
A new trait added for the barbarian file in order to give them pain resistance.

Barbarian Trait

## About The Pull Request

Giving barbarians a buff that sets them apart from the rest of the classes in the adventurer role, and their counterpart (Hunter) so that they're ultimately the warriors they're deemed to be.

## Why It's Good For The Game

The monks and other classes have higher strength than barbarians, but this is a class that is designed to be what a person chooses for a warrior class that embodies rage. Through this added trait, they have something that gives them a little support on this idea, especially giving them the idea that they will not need additional armor that separates them from being barbarians (steel plates and all of that bs).


JTGSZ:
Might want to check what you are doing buddy.

PR #641, opened 2024-05-07
Moves the bog gate and garrison near to the travel tile, leaves bog village outside of the gates

zakanater19:
## About The Pull Request

Garrison is connected by the gate roof, so bog guard can still patrol the bog village and bog without having to cheese the gates

## Why It's Good For The Game

Personally i think the bog village should be an unsafe place, bog guard can do what they can now but will probably fail, unless they have good leadership and support

What's left of the old bog gate

The new bog gate Z3
2 - Copy

The new bog gate Z5
2 - Copy (2)

Making sure bushes / grass spawn
2 - Copy (2) - Copy


Alibouk1:
I think I REALLY like this.

zakanater19:
Missed some rocks near the bog gate

Real-MAGNUM:
Looks great. What are your ideas for the rules within the town, now that the bog gate cant be used as reference? Replace after the boggate with inside the town?

zakanater19:
> Looks great. What are your ideas for the rules within the town, now that the bog gate cant be used as reference? Replace after the boggate with inside the town?

I'm leaning towards being RDM zone again, but with the bog guard there and having 20 pop they should be able to deal with murder hobos, especially considering adventurers will also gank murder hobos, but yeah im open to whatever ruleset

Real-MAGNUM:
Bogguards were a part of the problem. It's less allowing murder, and moreso the fact that people will walk in, climb a tree, then just start shooting anyone they want for like 20-40 minutes without admin intervention, potentially one shotting lotsa people.

zakanater19:
well imo the bog village should be a dangerous place, if you want safety you could go into the town, but again whatever rules administration ends up using is fine

Real-MAGNUM:
Testmerging. Rules are not changing. Only behind the gate are you protected.

TheManWithNoHands:
This will just be the final nail in the coffin for the "new" dwarf zone, and completly murder it. Same for town really.
Why should you build in rdm town, when next to rockhill its much safer? There is more then enough room around rockhill. Before the bog was not fully safe, with the ocational bandit/zombie horde breacking past the gate. Now there is no good reason to be in town over rockhill, expect looking for a challange. Like the townies that tried to build in town before bog change and got murderd every second round.

tl;dr: Macking the place that people use to relax and build/craft, into rdm zone be very bad. Dont fuck over the chill players.

-1

Real-MAGNUM:
this was a bad idea, nevermind.

dunsel-hordon:
A better idea would be a secondary gate inside the Rockhill area for the town guard to man. Control the flow of people from the bog better. But as I understand it the bog and town maps are supposed to be getting merged into one?

BlueMemesauce:
All merging would do is just remove the dumb travel thing and allow you to travel between zones instantly. As far as I know, there isn't actually any reason to have that when you can just put them on the same map idk why roguetown devs made it like tht

PR #637, opened 2024-05-07
20% less grain

NPC1314:
## About The Pull Request

* Adds alpha 20% to grain effect. Result is a slightly smoother grain.

## Why It's Good For The Game

Current grain effect too strong.

PR #642, opened 2024-05-07
Revert "20% less grain"

ChungusGamer666:
Reverts Blackstone-SS13/BLACKSTONE#637

shouldn't have done that since the pr directly touched icons instead of changing the alpha variable, very evil no good and icky


NPC1314:
The entire change was sliding the alpha channel value on the icons but sure, whatever.

Real-MAGNUM:
> The entire change was sliding the alpha channel value on the icons but sure, whatever.

He means changing it via code instead of the actual icon. He's right, it's icky. Please ask chungoid to help replace this in the way that he wants, he might be merciful

NPC1314:
I am aware what he meant, I am pointing out that its no more controversial than normalizing sound files instead of using vars to manage volume on a track by track basis. Context matters.

For the last part I can't believe you actually typed that. Thanks, I'm good.

PR #650, opened 2024-05-07
[FEATURE] Excidium: the bounty machine

Gyrandola:
## EXCIDIUM

From latin _massacre_, _slaughter_, the Excidium is a machine designed to allow anyone to put a bounty on someone else's head.

picture

**CREATING A BOUNTY**

Anyone can put a bounty on someone's head. To do so, one must interact with the Excidium and go through all the questions it will ask: name, reward, reason. With this information the Excidium announces the new bounty over SCOMM and updates the publicly available list of active bounties. Make sure to have enough funds in your SHYLOCK.

**COMPLETING A BOUNTY**

You decide to pursue a bounty. Very well, you only know the target's name and the reason why they are wanted dead, unless the creator was kind enough to give more information. Your job is to find the target and elegantly separate his head from the rest of his body, which you will present to the Excidium's maw for analysis. If the head is matches a bounty, the reward will be dispensed.

**EXCIDIUM LOCATION**
The Excidium is located in the central town plaza, in front of the guillotine.

**QUESTIONS**

Q: Why would the King allow this machine to exist? Murder is not nice.
A: The King benefits from sanctioned bounty hunting as a royal tithe of 10% is applied to each bounty reward.

Q: Can the machine be disabled?
A: Not as of now, but I intend to give the king or the steward this option.

image
                          _Example of bounties. The phrasing is generated randomly and filled with the real data._

image
_Example of the Excidium announcing the creation of a new bounty. It is also broadcasted to SCOM._
## Why It's Good For The Game

An indirect continuation of my mercenary rework PR, the Excidium incentivizes small-time contracts and mercenary work that anyone, alone or in a party, can pursue.


NGGJamie:
While the idea itself is neat, it removes a lot of incentive for player interaction, and kinda doesn't really fit in the world. For that last part, I mean that while it is a dark world, there is still some semblance of law that exists, and having a kill-for-hire machine in the middle of town that's just allowed to exist by the guard really doesn't mesh with the setting. Plus, it removes much of the risk from having to deal with mercenaries and trust them not to run off with your money, or double-cross you because someone else is paying more.

If anything of this vein were to be added, I would instead opt for some sort of a notice board where people can post up their requests and arrange meetings.

Gyrandola:
> I mean that while it is a dark world, there is still some semblance of law that exists, and having a kill-for-hire machine in the middle of town that's just allowed to exist by the guard really doesn't mesh with the setting. 

I was thinking of removing the possibility to put bounties on town/castle guards and other king's men.
Or perhaps place the machine in an area that is not necessarily under king's control, such as the anarchy zone. I do not want to be many of these things around, in fact probably just one.

>Plus, it removes much of the risk from having to deal with mercenaries and trust them not to run off with your money, or double-cross you because someone else is paying more.

That is true, but on the other hand most mercenary bands tend to get hired "in bulk" by either the King or the Lord of Heartfelt since they are the richest employers. I rarely, if at all, see people put a bounty on someone elses head unless, again, they are the king using the Outlaw command.

I believe this machine encourages smaller and less round-long contracts that anyone can pursue, not just mercenaries. It does take away some rp situations, as you said, but it also creates others. Take for example a situation where multiple parties are hunting for the same person and do not want to share the reward.

> If anything of this vein were to be added, I would instead opt for some sort of a notice board where people can post up their requests and arrange meetings.

I suppose that can be done, but for now I wish to mantain my original direction unless my hand is forced.

Wintermote:
Major issue I've run into as merc is that I'm just never around when someone wants to hire me, and when someone does hire me they feel like they can just pay the money on their own time (45 minutes after the job is done). It's just in general very hard to actually get merc jobs and then on top of that get paid for them in a timeframe that's relevant.

One thing I'd like to see from an automatic escrow service, is a printed job contract. A good example of this were the Morag Tong assassin's guild, honorable writs of execution from Morrowind:

>The afore-mentioned personage has been marked for honorable execution in accordance to the lawful tradition and practice of the Morag Tong Guild. The Bearer of this non-disputable document has official sanctioned license to kill the afore-mentioned personage.

Gyrandola:
> One thing I'd like to see from an automatic escrow service, is a printed job contract. A good example of this were the Morag Tong assassin's guild, honorable writs of execution from Morrowind:

A job contract of that kind would imply that a contract can only be taken by one person(or by extension a group) at a time.

Right now the machine only works as a notice board and automatic reward giver, this means multiple people can independently decide to pursue a bounty and whoever delivers the head first wins.

I like the competitive aspect of the latter, but nothing is decided yet.

Sabastian999:
this could also be applied to idem request

Real-MAGNUM:
Testmerging.

TheManWithNoHands:
> this could also be applied to item request

If you did that this would just fully turn into a quest board mashine.
Im more of the opinion to make an actual adventures guild in the ground level of the inn.
Would fit more the setting and that room is semi unused anyway.

Gyrandola:
fuck  rebases im reopening

PR #652, opened 2024-05-07
unfucks prisoner and rebalances his stats

bobjension:
I'm not sure whoever was smoking, but giving the prisoner role infinite slots and -4 strength is insane


YosemiteYam:
I thought they were infinite slots due to people with a bombed PQ below -30 will only be able to spawn as prisoners, and the shitty stats are there to not reward them for ending up as prisoners after shittery.

bobjension:
who is gonna have -30 PQ reasonable? they would be probably be banned before that happens lol

CVSthePharmacy:
> who is gonna have -30 PQ reasonable? they would be probably be banned before that happens lol

zizo bomb

PR #662, opened 2024-05-08
Shield speed debuff

panvxv:
## About The Pull Request
Decreases your speed if you have a shield in hand. So weapons that speed matters like knives, rapier and sabre won't be able to get extra bonuses from user's speed.
Also to make some more differences between bigger and smaller shields (incoming buckler!).
Speed debuff is small enough so it won't effect your walking speed but if you really want to walk faster, just put it on your back!

## Why It's Good For The Game
Balance!

### DIFF
--- a/code/game/objects/items/rogueweapons/shields.dm
+++ b/code/game/objects/items/rogueweapons/shields.dm
@@ -99,6 +99,8 @@
 	coverage = 70
 	parrysound = list('sound/combat/parry/shield/towershield (1).ogg','sound/combat/parry/shield/towershield (2).ogg','sound/combat/parry/shield/towershield (3).ogg')
 	max_integrity = 200
+	var/shielding
+	shielding = FALSE
 
 /obj/item/rogueweapon/shield/tower/getonmobprop(tag)
 	. = ..()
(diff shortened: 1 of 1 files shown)


hivemuscle:
bow/xbow agent detected, pr rejected

CVSthePharmacy:
shields don't have constant damage blocking anymore, you have to have it on hand with block mode selected AND CHARGED to actually block properly, so the speed debuff will make shields much more shit than they already are. get with the times old man, they aren't as good as before

panvxv:
> shields don't have constant damage blocking anymore, you have to have it on hand with block mode selected AND CHARGED to actually block properly, so the speed debuff will make shields much more shit than they already are. get with the times old man, they aren't as good as before

Shields are still one of the best items in the game, this only effects tower shields anyway.

Wintermote:
Feels more like the debuff should be on the weapon you don't want used with a shield, rather than on all shields at all times

panvxv:
> Feels more like the debuff should be on the weapon you don't want used with a shield, rather than on all or some shields at all times
> 
> Outside the context of the game (e.g. not necessarily relevant for gameplay but I think it's worth adding), the serious downside for rapiers historically was them being utter shit on a battlefield; they're a civilian self defense / arranged duel weapon, not a battlefield weapon. They bend and break more easily, lack reach of most other battlefield weapons, are so light it makes binding and parrying heavier weapons more difficult, and aren't good at piercing through armor.
> 
> partially the same reason bucklers or daggers were used with rapiers - small, light, easy to carry, perfect for civilian use.

I will nerf rapier, I know already it's integrity will be reduced from 150 to 75. But for more changes I need to think them through. I am always trying to make changes as much realistic as possible when trying to balance.

Real-MAGNUM:
Good attempt but we already have a pretty decent nerf for shields atm.

ChungusGamer666:
No more shield nerfs until directionality is added, this is my official ruling. We are absolutely sidestepping the biggest issue with shields...

PR #663, opened 2024-05-08
crossbows gets AP and bows get their damage buffed

bobjension:
god made man, who were equal. 
man made crossbow so man could be equal.

bows got 20 damage before, now 35, which is DOGSHIT a sword did more damage and you can actually repeatedly attack with it

### DIFF
--- a/code/game/objects/items/rogueweapons/ranged/ammo.dm
+++ b/code/game/objects/items/rogueweapons/ranged/ammo.dm
@@ -50,7 +50,7 @@
 	caliber = "arrow"
 	icon = 'icons/roguetown/weapons/ammo.dmi'
 	icon_state = "arrow"
-	force = 20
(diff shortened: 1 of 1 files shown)


bobjension:
it compiles YAY

CVSthePharmacy:
yay now I won't be stopped by a gambeson when I try to shoot a bandit

efzapa:
waiter!! waiter!!

more one hit weapons please!!

efzapa:
> > waiter!! waiter!!
> > 
> > more one hit weapons please!!
> 
> But every weapon can one hit. Even a stone can.

problem being you can't parry a body and NOW armor is even more useless than it was

also the claim that weapons can one hit is really only true if you have like 15 STR and crit on your first hit against someone with no gorget

which is really just cherrypicking, rarely in reality will you get one hit and when you do, it's usually a skill issue like not having a visor down

PR #667, opened 2024-05-08
Makes prisoner roundstart only to fix them latejoining on the boat outside cells

zakanater19:
## About The Pull Request

Temporary fix for me not being able to figure out how to make prisoners spawn in the cells latejoin

## Why It's Good For The Game

Prisoners no longer latejoin outside of cells

### DIFF
--- a/code/modules/jobs/job_types/roguetown/peasants/prisoner.dm
+++ b/code/modules/jobs/job_types/roguetown/peasants/prisoner.dm
@@ -3,7 +3,7 @@
 	flag = GRAVEDIGGER
 	department_flag = PEASANTS
 	faction = "Station"
-	total_positions = -1
+	total_positions = 0
 	spawn_positions = -1
 
 	allowed_sexes = list(MALE, FEMALE)


TheManWithNoHands:
Maybe look how the adventures and pilgrims can spawn not in the boat as late join.
Thats a role restricted spawn point after all.

zakanater19:
> Maybe look how the adventures and pilgrims can spawn not in the boat as late join. Thats a role restricted spawn point after all.

Yeah i tried using it as a reference but using the adventuerlatejoin spawns and overriding the jobrole to prisoner still didn't work even after like four attempts at tweaking it, and even making a new marker didn't work

BlueMemesauce:
Adventurers spawn in the forest along the edges of the map which makes sense. Prisoners randomly appearing in a cell wouldn't make much sense

thebigtony:
Spawning in a cell would make more sense than spawning on the boat. But I think this is a good PR.

Real-MAGNUM:
Testmerging. You probably need to create a new spawn marker for prisoners in specific.

CVSthePharmacy:
people with -30 pq can't spawn anymore, this is a shadowban now, revert this pr

BlueMemesauce:
Only if admins give you negative pq

CVSthePharmacy:
> Only if admins give you negative pq, which shouldn't be happening

we just disproved what you said, now revert it as it shadowbans people

BlueMemesauce:
The admins can also just straight up ban people so

PR #668, opened 2024-05-08
BUCKLER!!

panvxv:
## About The Pull Request
Adds buckler, a recipe for it and new mechanic just for it.
Buckler is a small shield held in one hand, it doesn't protect again incoming range attacks but sure will work greatly with one handed weapon!

## Why It's Good For The Game
New kind of shield and more options for smiths and fighters!


Real-MAGNUM:
Testmerging

PR #673, opened 2024-05-09
More bog ambushes 

Osmos123:
## About The Pull Request
Adds in the volves, skeletons, and minotaur's into the bog ambush list.

## Why It's Good For The Game
Suffer and pain adventurers and pilgrims.

image


pismorizzo:
Was this tested or fought at all? The minotaurs have 300 hp and hit for _45-75_ damage each and spawn in groups of 2. For comparison a vulf has 120 hp and hits for only 19-29. Additionally it fully decaps and delimbs a body. This things are MONSTEROULY OP a gorup of adventurers is basically just dead if they get ambushed by these and cant outrun, since they will be dead before even half healthing a single 1.

dunsel-hordon:
Bad merge either nerf em or take this out

PR #683, opened 2024-05-09
Archivist goon

panvxv:
## About The Pull Request
Moved archivist to nobility (basement noble), made him pacifist and legendary at every skill that isn't crafting (for now as I need to make new trait which will prevent him from crafting).
This should make him more interesting as a man/woman of skill to share with others for some coin.

## Why It's Good For The Game
FUNNY! Unique role for just for Ye!


CVSthePharmacy:
give him max weaponsmithing and armorsmithing so he can teach, oh right you can't because YOU CAN'T LEVEL UP SMITHING ANYMORE

BigBigga:
based magnum

PR #710, opened 2024-05-10
Adds Max PQ framework, adds the Lunatic role.

efzapa:
## About The Pull Request

Having -50 PQ now locks you to being a Lunatic which marks you as a griefer IC so people can valid RDM you.

Also sets Prisoner minimum PQ to -49

## Why It's Good For The Game

Point of lunatic OOC wise is to force people to dig themselves out of their -50 tide hole manually and roleplay for commendations like a normal person
Or just you know, keep getting RDM'd

Most of the reason I made this is because like half the people that roll prisoner wind up getting released and it's meant to be a punishment role most of the time

### DIFF
--- a/code/__DEFINES/jobs.dm
+++ b/code/__DEFINES/jobs.dm
@@ -143,6 +143,7 @@
 #define ADVENTURER	(1<<9)
 #define COOK		(1<<10)
 #define GRABBER		(1<<11)
+#define LUNATIC		(1<<12)
 
 #define YOUNGFOLK		(1<<5)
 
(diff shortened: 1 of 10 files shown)


BlueMemesauce:
I think max pq is a bad idea, we shouldn't encourage players to be shitters

efzapa:
> I think max pq is a bad idea, we shouldn't encourage players to be shitters

Unless someone is TRYING to get their stats tanked and RDM'd every round, i don't see this happening
In which case they will likely get banned just as easily

PR #716, opened 2024-05-10
Random map loader 

FlickTee:
## About The Pull Request

This hijacks and adds functionality to two things; the in-suite .dmm loader and the landmark effect, to allow mappers to create a .dm file linking any number of .dmm maps, place a landmark anywhere on the map, and load a random dmm out of that list to spawn the map in at round start. This allows random dungeons/landmarks, ruins building variants, or even entire randomized towns, etcetera. It should properly account for areas, power, lighting, effects, z-levels, and so on. What it does NOT do is delete objects that are already there on round start nor does it randomize the placement of the markers- maps will always spawn at the mark, which is at the bottom left of the loaded .dmm (ie. the marker will always start from 1,1,1 in the mini dmms).

## Why It's Good For The Game

Random variants on the map allows mappers several benefits - editing their dungeon/etc does not require a full map edit (thus keeping accidental overwrites to a minimum), as well as more variation on each round.


Real-MAGNUM:
Thanks for doing this. I was going to need this for dungeons

PR #722, opened 2024-05-10
Necromancer fix

Ambrose997:
## About The Pull Request

Zombies aren't the allies of Necromancers, so I see no reason why the necromancers spells should heal them as opposed to stunning them and doing damage. As it stands currently necromancers are basically helpless against zombies aside from eye-bite so it seems incredibly stupid to not rectify this.

## Why It's Good For The Game

Necromancers are now not ez zombie food.


rusher210:
while i thanks necros aren't zombie food anymore i do find cringe that NECROmancers, the ones that summon the DEAD don't interact with zombies and only skeletons, zombies without flesh

DirtTastesBad:
Could add a sort of "Imprint" spell that keeps zombies from attacking you, and gives them an objective to followbyour orders. Could even turn them into rotmen or something like that, but that might be going too far balance wise

DirtTastesBad:
But if that option isn't on the table, removing the undead healing is probay for the best

TheManWithNoHands:
Honestly, being able to gain controll over undead sounds fun. Problem though. The autobite targets EVERYTHNG. It targets every mob, also other zombies. For a spell to make zombies friendly to the necromancer, you would need to dable in the autobite thing. Otherwise you could only tame zombies you cant get close to.

ChungusGamer666:
no more half measures, friendly fire nuke PQ soonish (?)

Ambrose997:
> no more half measures, friendly fire nuke PQ soonish (?)

I've had admins confirm they arent allies and no one intends for them to be. It's not a half measure.

ChungusGamer666:
to be decided with staff ig

SELFHELL:
Theoretically zombies SHOULD be ganging up with Necros, but player freedom ect

Real-MAGNUM:
I'd rather have a friendly fire PQ removal of -0.2, and only for specific factions that dont turn on eachother often. (bandits, vampires, zombies) 

Undead overall should be healed by unholy magic regardless of faction. changing this will create unpredictable inconsistencies in both magic gameplay and its lore.

It's bad for necromancers, but necromancers are already annoying for everyone else. Let the dead they linger around be annoying too. Necromancers arent automatically enemies, people have misconceptions about that. Re-closing.

PR #727, opened 2024-05-10
Map Changes

JTGSZ:
I was DM'd to PR these map changes.
All credits go towards Sanghelli and song, M.

I looked it over, but the person who does the most mapping should probably review this.


JTGSZ:
handed to the person who knows the current map the best rn

PR #730, opened 2024-05-10
Adds option to change FPS

Dezupher:
## About The Pull Request
Adds a verb to change your client FPS in the options tab.

## Why It's Good For The Game
There's no way to change your FPS right now.  #524 set the default to 100, however if you connected to the server before this pr was merged your FPS was set and saved to the previous default of 20 and you were stuck with it.


Osmos123:
HUGE

PR #733, opened 2024-05-11
Adds a stockpile and submssion hole to the bog

gnomedeplume:
## About The Pull Request

* Adds a stockpile and submission hole to the bog at the entrance of the inn near the shylock.
* Makes it so literacy is not checked when using stockpile and submission hole

## Why It's Good For The Game

This is specifically so that soilsons and soilbrides can sell their goods without having to walk all the way to town. It will also make is so that adventurers and bog innkeepers and bog smiths can more easily buy and sell goods, which I consider to be beneficial.

Right now, as a soilson/bride, your best strategy is to put everything into bags/carts, drag it all the way to town, and then just sell everything at the submission hole anyway because the innkeeper probably won't pay you more than the submission hole and the merchant could be missing / dead. (That's what happened in my previous game in any case.) It just adds busy work without actually adding much opportunity to role play.

## Testing

Spawned as soilbride, sold some grain, bought some wood.

image


gnomedeplume:
Moved to https://github.com/Blackstone-SS13/BLACKSTONE/pull/734

PR #737, opened 2024-05-11
Stockpile adjustments

Osmos123:
## About The Pull Request
"Fixes" the stone resource to actually take in the small rocks. Adds in slightly more iron in stock at the beginning as a buffer for smiths. Makes hides and fur a little more lucrative for adventurers to collect. Meat, eggs, poultry, and apples can be deposited also. Import/export quantity has been normalized. Hopefully taverns/inns/butchers can take advantage of this now.

Reduces to the profit margins from gold ore -> gold ingot (100 sell price) -> three gold rings smith craft (135 sell price total)

Feel free to let me know which prices should be adjusted, or any of that sort.

## Why It's Good For The Game
Attempts to expand on the economic system.

image
image
image


code/modules/roguetown/roguestock/stockpile.dm
```diff
@@ -43,56 +43,92 @@
...
 	name = "Hide"
 	desc = "Stripped hide from animals."
 	item_type = /obj/item/natural/hide
 	held_items = 0
-	payout_price = 3
-	withdraw_price = 10
-	export_price = 7
-	importexport_amt = 10
+	payout_price = 10
+	withdraw_price = 14
```
gnomedeplume, line 67: How would you feel about making the withdraw price consistently double the payout price? Right now the ratio is all over the place. I think having a consistent ratio would be easier to understand and would make traders doing price arbitrage (that is, buying higher than payout, selling lower than withdraw) more viable as a job. If you don't want to make the ratio 2:1, can we at least make it consistent? Maybe 3:2?

code/modules/roguetown/roguestock/stockpile.dm
```diff
@@ -43,56 +43,92 @@
...
-	held_items = 2
-	payout_price = 1
-	withdraw_price = 5
-	export_price = 5
-	importexport_amt = 15
+	held_items = 5
+	payout_price = 2
+	withdraw_price = 4
+	export_price = 4
+	importexport_amt = 10
```
gnomedeplume, line 59: Agree with this price increase. Can we also add fiber at payout_price = 1? Fiber is incredibly useful in crafting.

code/modules/roguetown/roguestock/stockpile.dm
```diff
@@ -43,56 +43,92 @@
...
+	name = "Bird Meat"
+	desc = "Edible flesh harvested from birds."
+	item_type = /obj/item/reagent_containers/food/snacks/rogue/meat/poultry
+	payout_price = 5
+	export_price = 8
+	withdraw_price = 8
+	importexport_amt = 5
+
+/datum/roguestock/stockpile/poultry
+	name = "Egg"
```
gnomedeplume, line 128: Can we add either milk or cheese and butter?

code/modules/roguetown/roguestock/stockpile.dm
```diff
@@ -26,13 +26,13 @@
...
 	export_price = 6
-	importexport_amt = 20
+	importexport_amt = 10
 
 /datum/roguestock/stockpile/iron
 	name = "Raw Iron"
 	desc = "Chunks of iron used for smithing."
 	item_type = /obj/item/rogueore/iron
-	held_items = 4
+	held_items = 12
```
gnomedeplume, line 35: I think this is a bit high. It could make mining less enticing as a profession. I think 6 would be a more reasonable target here.

I think making mining more convenient at round start, such as by adding a bunch of picks to the map, would be a better fix here. And maybe reworking the drop rate of raw iron in the mine so that it's more consistent to get iron.

gnomedeplume (commented):
Really like these changes overall. I was planning on working on something similar soon. Left some comments.

code/modules/roguetown/roguestock/stockpile.dm
gnomedeplume: Another thought... I still think the 2:1 suggestion makes sense right now. However, in the future I have an idea I may want to implement. I'd like to have stockpiles have their own distinct piles of goods and moving goods between stockpiles automatically would cost mammon. In that case, buying and selling to the same stockpile would have a much smaller margin. My thought would be a 20% margin but the steward could adjust that margin.

Real-MAGNUM:
Tell me when you're ready for testmerge. Also consider increasing the time it takes for food to spoil.

Osmos123:
Increased food spoilage time and other adjustments requested. It is ready for test merge.

gnomedeplume:
Thanks for making some of my suggested changes! Glad this is moving forward ^_^

Real-MAGNUM:
testmerging.

Real-MAGNUM:
untestmerging, something conflicted.

Osmos123:
Fixed to include the changes of the #782 and ready for test merge again.

code/modules/roguetown/roguestock/stockpile.dm
```diff
@@ -47,63 +47,140 @@
...
+	withdraw_price = 13
+	transport_fee = 3
+	export_price = 13
+	importexport_amt = 5
+
+/datum/roguestock/stockpile/cheese
+	name = "Cheese"
+	desc = "The product of milk and salt."
+	item_type = /obj/item/reagent_containers/food/snacks/rogue/cheese
 	held_items = list(0, 4)
```
gnomedeplume, line 181: Not a big deal, but this changes from having 4 grain in the stockpile to 4 cheese in the stockpile. 

Sorry for the conflicts. >_<

Osmos123: All good, switched for grain to have 4 in stockpile.

Real-MAGNUM:
Testmerging.

Real-MAGNUM:
Havent heard any complaints or bugs.

PR #741, opened 2024-05-11
readingGODS Readd reading requirement for submission hole and stockpile

bobjension:
filthy peasants who can't read don't deserve the luxury of opening the stockpile or the submission hole

(beggars can open the stockpile, which is weird)


Real-MAGNUM:
We JUST did the opposite because of people like soilsons being unable to operate them. I think these machines are simple enough to have icons or physical representation instead of reading, or even word of mouth to remember which buttons produce which item.

PR #748, opened 2024-05-11
Spellbooks

NGGJamie:
## About The Pull Request

Adds spellbooks to the game that can teach the player any spell marked as "Learnable"

Note, I tossed learnable on a bunch of spells, including ones that don't work just to see how/if they'd function. The current list of spells flagged as learnable are not final, many need to be removed.

TODO: There is no natural way to create these books. They must be admin-spawned and then have the pick_spell proccalled onto them.

TODO: Implement rarities for spells properly, and find a balance where skill can relate to level of spell that is creatable.

## Why It's Good For The Game

Magic is dogwater right now, we need variety.


Osmos123:
Could you also include trait books also such as giving heavy armor training or maybe even +1 stat increases? Maybe even unholy books that could give one no stamina cost or bleeding.

NGGJamie:
On hold until the anti-ban evasion project I'm building is in production.

PR #755, opened 2024-05-12
Adventurer weapon skill standardization

Osmos123:
## About The Pull Request
Normalizes the relevant weapon skill for each adventurer to be skilled (level 3) in consideration to the 75 cap and virtually infinite number of adventurer scourge. This can always be adjusted when adventurer have finite caps.

The pecking order of superiority will be Knights (master level) -> Castle Guards (expert level) -> Templars (expert level) -> mercenaries german ones and desert riders (expert level) -> Town Guards (skilled-expert level) -> adventurers (skilled level) -> Bog Guards (now skilled level) -> pilgrims (novice-average level) -> everyone else. 

Avoided touching the uniques such as witch hunters or lord of heartfelt.

## Why It's Good For The Game
Makes it challenging for adventurers to murderbone the limited amount of slots of mercs/castle guards/knights. Increases the need for sparring to train up (will need to figure out to include a very rare/limited legendary adventurer sorta like a frail blademaster role).

Hopefully encourages people to fill these limited roles rather than going straight to adventurer also.


YosemiteYam:
Good PR, though honestly Bog guard are very gimped with the new training and have noticed no one bother going bog because they will always be outclassed by literally everyone that isn't a civilian. Maybe atleast put their sword skill to skilled same as adventurers and give them average in polearms and maces (They do not have polearm skill at all even though they have spears in the armory and they usually roll weak in maces instead of average)

Osmos123:
> Good PR, though honestly Bog guard are very gimped with the new training and have noticed no one bother going bog because they will always be outclassed by literally everyone that isn't a civilian. Maybe atleast put their sword skill to skilled same as adventurers and give them average in polearms and maces (They do not have polearm skill at all even though they have spears in the armory and they usually roll weak in maces instead of average)

Gave bogguards skilled for axesmaces, swords, polearms. Reduced slots from 20 to 10 (I think there was the desire to reduce their slots in other pr's that did not go through).

GreenHoodiedEthan:
I'm a bit sad that Warrior loses it's skilled profs in Axes & Maces / Polearms and some of the other classes that had expert profs. But I suppose if it's to prevent the adventurer rush it has to be done. Some of the villager classes are gaining weapon skill on par with the adventurers which is kind of funny to think about.

SELFHELL:
Peak PR. Death to adventures

PR #763, opened 2024-05-12
Goblin Race!

panvxv:
## About The Pull Request
Goblin race being playable race from roundstart!
As for now players won't spawn as there isn't a role for goblins. But they can look at the little goblin on during lobby time!

## Why It's Good For The Game
GOBO!


assortedbeads:
Tempted by that which we cannot reach

SELFHELL:
This is extremely reduntant to merge when theres nothing to even do with the race.

Real-MAGNUM:
Having it means it's easier for things to be done with it. It will be excluded from the character creation menu via config.

EnzoSeptum:
You make me happiest man alive

PR #766, opened 2024-05-12
Rockhill is vulnerable to zombies when the Priest dies

ChungusGamer666:
if the priest dies, zombies can rise in rockhill just as if it was in the forest


BlueMemesauce:
should probably allow a new priest to be made then
maybe the king can announce a new priest

PR #767, opened 2024-05-12
Adds an archivist basement corridor to the garrison office and fixes the archivist spawn on the actual map this time

Warfan1815:
## About The Pull Request

I forgot we had moved to blackstone.dmm, so that last fix isn't going to work

I've also added a basement corridor from the archivist office to the garrison office

image
image
image
image

## Why It's Good For The Game

A corridor is good because it slightly undermines the defendability of the keep, moreso than skybridges &c (on which I promise to fix in a PR soon tm) but also allows the archivist in normal play to have a front-facing place to talk to people.


Real-MAGNUM:
@ChungusGamer666 Please stop merging mapping PRs without checking to see if they used mapmerger2. It keeps breaking the entire map.

PR #775, opened 2024-05-13
Less crits

Bakeun0:
## About The Pull Request

You cannot critically hit a limb unless it is damaged by at least half its maximum damage cap (At 10 CON, this should be 100 for head/chest and 50 for limbs). this is checked AFTER damage is applied, so crits can still happen in a single hit if the attack is strong enough.

## Why It's Good For The Game

Makes a Fair and Balanced mechanic less boring.


EnzoSeptum:
Combat will become click simulator if this shit gets merged.

Bakeun0:
I can't imagine anyone enjoying random crits the way they are now but it seems like this isn't the way to improve them. Closed

PR #782, opened 2024-05-13
Make stockpile local

gnomedeplume:
## About The Pull Request

* Make goods in the stockpile local
* Goods can be withdrawn from remote stockpile for a fee
* Stockpile handles submission and withdrawing goods
* Remove submission hole
* Replace stockpiles in shops with "vomitorium" which is a withdraw-only machine
* Update steward machine to use goods from both stockpiles

## Why It's Good For The Game

In a previous PR I added a submission hole near the farm to make sure grain and other food items get distributed through the economy. Some folks complained that this discouraged people from traveling from the town to the bog. By adding a transport fee, it gives people who want to ship goods back and forth an incentive as they can take that transport fee for themselves.

## Possible future improvements

* Instead of transport fees that magically disappear, store them in an escrow account. Using transport degrades the transport system. The transport system can be repaired with cogs. Repairing gives a coin bounty from the escrow account.
* Add a local transport fee to use the vomitorium? Would have to be a very small amount, which would probably require bulk withdrawal. 

## Testing
image
image
image
image
image
image


TheManWithNoHands:
Making the Stockpiler less RP killing is something I'm fully on board with.

Having people literally pay to use the less RP option is good, but there is already a problem with the prices being too high.
So the fee prices might need some testing.

PR #784, opened 2024-05-13
bundle changes

lovegreenstuff:
## About The Pull Request
does three things mainly
1) bundling behavior is now part of natural and natural/bundle, so it should be easier for people to make new bundle types. only coders care about this though.
2) crafting can now use bundles. you don't have to unbundle everything anymore
3) right click on fibers to collect all of them from the tile, soilson mains rejoice

## Why It's Good For The Game
i don't remember why i started working on this pr
it took me like 4 days of testing and i kind of lost the plot

https://github.com/Blackstone-SS13/BLACKSTONE/assets/59631103/274ab0ed-b3fb-4f0c-ad48-d362d10537c7

### DIFF
--- a/code/datums/components/crafting/crafting.dm
+++ b/code/datums/components/crafting/crafting.dm
@@ -102,6 +102,9 @@
 		if(istype(I, /obj/item/stack))
 			var/obj/item/stack/S = I
 			.["other"][I.type] += S.amount
+		else if(istype(I, /obj/item/natural/bundle))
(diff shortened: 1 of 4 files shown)


gnomedeplume:
Would love to be able to insert bundles into the stockpile at some point. Probably should go in a separate PR, especially since I'm actively reworking the stockpile right now.

ophaq:
Is this supposed to work with sticks because it currently doesn't and sticks would be nice to have.

PR #787, opened 2024-05-13
Salt cave edits and lava cave edits

Real-MAGNUM:
Changes the entrances to both caves slightly, as pinned in mentor help


dunsel-hordon:
nice changes

PR #795, opened 2024-05-14
Replaces the bathhouse under the Inn with more Inn room and adds quest boards

TheManWithNoHands:
## About The Pull Request
* Gets rid of the old ERP are and replaces it with an RP area.
* Adds some Pin boards in the lower part of the tavern.
* changes the direction of the Inn staircase
* Changes the Inn storage a bit.

## Why It's Good For The Game
Getting rid of any trace of ERP is good.
Also gives a place where people can more easily post quests, to the area where people already do that.

https://imgur.com/a/Ah9Ip3W


TheManWithNoHands:
Ok, now it should be fixed

YosemiteYam:
I don't see the point in giving the inn another floor of tables and counter to deal with, the ground floor is more than enough to deal with customers and theres no incentive for a single innkeeper to go back and forth between the lower and upper level. All this will do will give a free clubhouse to whichever band of adventurers or heavily armed group takes it over and bullies the innkeeper to not go "I own this tavern" without being one shot to the head. It also doesn't fix the problem the lower level of the inn constitutes = there is no point to it. Like you can just put the quest notice board on the ground floor of the inn, why add another level for it. 

I get that the current erp pools are dead space but changing it to just another level of tables and chairs for people to chill at for free is also dead space. I would much rather see an actual use come from this space that would benefit the barkeeper/innkeeper like moving the arena here, having a workplace of sorts, a mushroom farm, a drug den with hookahs or drug making supplies etc besides just turning the pools into another floor of a tavern that the innkeeper won't even be able to keep an eye on consistently.

Hmeister-real:
Maybe a little medieval gambling basement would be fun? I would love to do medieval blackjack and lose all of my Mammons. also Craps would work with the die we already have, just needs to be mechanically enforced. Roullete would be fun and all, but it's not really medieval

TheManWithNoHands:
> I don't see the point in giving the inn another floor of tables and counter to deal with, the ground floor is more than enough to deal with customers and theres no incentive for a single innkeeper to go back and forth between the lower and upper level. All this will do will give a free clubhouse to whichever band of adventurers or heavily armed group takes it over and bullies the innkeeper to not go "I own this tavern" without being one shot to the head. It also doesn't fix the problem the lower level of the inn constitutes = there is no point to it. Like you can just put the quest notice board on the ground floor of the inn, why add another level for it.
> 
> I get that the current erp pools are dead space but changing it to just another level of tables and chairs for people to chill at for free is also dead space. I would much rather see an actual use come from this space that would benefit the barkeeper/innkeeper like moving the arena here, having a workplace of sorts, a mushroom farm, a drug den with hookahs or drug making supplies etc besides just turning the pools into another floor of a tavern that the innkeeper won't even be able to keep an eye on consistently.


When you map it.

TheManWithNoHands:
> Maybe a little medieval gambling basement would be fun? I would love to do medieval blackjack and lose all of my Mammons. also Craps would work with the die we already have, just needs to be mechanically enforced. Roullete would be fun and all, but it's not really medieval

I did add more gambling desks, yes

gnomedeplume:
Can you add a millstone and a well to the inn?

TheManWithNoHands:
> Can you add a millstone and a well to the inn?

The inn already has a milstone, and but i did add a well so the inn would have fresh water again after its previous source (the kanal water) got popyfied.

BlueMemesauce:
I use the bathhouse to clean myself sometimes
it also gives a positive moodlet if you take your clothes off

TheManWithNoHands:
Merge conflicts because map updates :P

PR #797, opened 2024-05-14
Fixes the one tap merchant windows

zakanater19:
## About The Pull Request

Gives the merchant windows that take some effort to break down

## Why It's Good For The Game

People will stop rushing the easy smash and grab window that is the merchants bane currently


CVSthePharmacy:
the stone of zizo in my pocket:

MickDick1:
This is really important. Right now, it's always some stupid rush to get material and building walls right away.
Not doing that means you get killed. 100% of the time. It makes for every-single-round busy rush work for the sake of it. If I had it my way I would just replace these with walls.

PR #799, opened 2024-05-14
Whip balance changes

Hmeister-real:
## About The Pull Request

 Whips are only craftable by skincrafters, which is mostly adventurers and pilgrims, specifically Hunters and rangers, and these two almost never have enough strength to actually use the whip. Also makes it so the whip needs a stone to craft, since the description says it needs a stone (repenta en should need to be forged but i dunno how to do that)

## Why It's Good For The Game

 Allows whip to be used by the people who craft it, also i never see it used since most don't know it exists, hopefully now that it can actually be used by skincrafters it will be made and get spread around after they die.


OrbitalParadise:
Why need a stone to make a whip?

PR #800, opened 2024-05-14
Small Vaquero Changes

GreatRuna:
## About The Pull Request

Vaquero is given a 33% chance to have a orange cloak, and a 33% chance to have a black cloak. Vaquero gains a guitar, a small handful of money, and a dagger.

## Why It's Good For The Game

At the moment, Vaquero does not have any money on spawn, despite the station of their clothes and their steed. With a pitiful amount of money, this change gives them enough to get off their feet and matches the other adventurer classes in that regard. 

The instrument is given because Vaquero has a basic music knowledge, and no way to use it. It would make sense a nomadic traveler would keep one, and gives them a chance to use the music skill they are granted.

The dagger is due to the fact duelists commonly employ parrying daggers in a fight, as well as the general benefits of having a knife on an adventurer, which most, if not all, seem to have already.

The cloak color variation is due to the fact Vaquero is a tiefling, so it blends a little, and gives slight cosmetic variety for no other purpose than vanity a la bards having a chance to have a red cloak instead of blue. Orange was chosen to match the vibrancy of red, and black was chosen because it matches the idea of what vaquero is themed.


CVSthePharmacy:
oh look a QOL, speedmerge

PR #801, opened 2024-05-14
Templar Tabards with Faith Coloring

GreatRuna:

### DIFF
--- a/code/modules/clothing/rogueclothes/cloaks.dm
+++ b/code/modules/clothing/rogueclothes/cloaks.dm
@@ -136,6 +136,31 @@
 	color = CLOTHING_RED
 	detail_color = CLOTHING_WHITE
 
+/obj/item/clothing/cloak/tabard/crusaderastrata
+	detail_tag = "_psy"
+	color = CLOTHING_ORANGE
+	detail_color = CLOTHING_YELLOW
+
+/obj/item/clothing/cloak/tabard/crusaderdendor
+	detail_tag = "_psy"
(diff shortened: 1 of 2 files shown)


GreatRuna:
may change the code so the tabard is split between the two main color, since that would look pretty nice

code/modules/clothing/rogueclothes/cloaks.dm
```diff
@@ -136,6 +136,31 @@
 	color = CLOTHING_RED
 	detail_color = CLOTHING_WHITE
 
+/obj/item/clothing/cloak/tabard/crusaderastrata
```
ChungusGamer666, line 139: bad use of inheritance
create /obj/item/clothing/cloak/tabard/crusader and put detail_tag = "_psy" there

ChungusGamer666 (changes requested):

code/modules/clothing/rogueclothes/cloaks.dm
GreatRuna: I'm a little confused, this is my first time coding for byond

I can't find /obj/item/clothing/cloak/tabard/crusader in the directory anywhere, and when I ran this locally it worked fine

GreatRuna: I got some clarification, thank you

GreatRuna: resolving because i think i fixed this

PR #807, opened 2024-05-14
Crossbow and bow balance

panvxv:
## About The Pull Request
No more scaling with PER, as having more than 100dmg meant it went through armor and crit people.

## Why It's Good For The Game
Armor should protect against range attacks.

### DIFF
--- a/code/game/objects/items/rogueweapons/ranged/ammo.dm
+++ b/code/game/objects/items/rogueweapons/ranged/ammo.dm
@@ -18,7 +18,7 @@
 
 /obj/projectile/bullet/reusable/bolt
 	name = "bolt"
-	damage = 65
+	damage = 80
 	damage_type = BRUTE
 	icon = 'icons/roguetown/weapons/ammo.dmi'
 	icon_state = "bolt_proj"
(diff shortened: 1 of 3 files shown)


Ambrose997:
Might as well delete the perception stat.

Deroh-Ra:
Crossbows by design are meant to pierce armor, bows shouldn't be able to pin armor though

Hmeister-real:
bows are gonna be fucking worthless, at least buff them to do something unique, like maybe they do a a ton of damage when you remove them and when they fall off

pismorizzo:
Bow damage was already buffed and currently does more damage than crossbows when skilled, dont understand the complaints against them. 
I do agree this makes perception kinda weak and perhaps should simply reduce or cut the scaling in half like was done with strength instead of removing all scaling outright.

YosemiteYam:
The ranged cabal needs to stop crying over not being able to one shot people

CVSthePharmacy:
> The ranged cabal needs to stop crying over not being able to one shot people

you should be banned from the github

OrbitalParadise:
> The ranged cabal needs to stop crying over not being able to one shot people

All weapons can one shot get over yourself

Real-MAGNUM:
This is a weird idea to fix a problem we all know exists. Remove scaling from crossbow only, it's a simpler weapon to use, has more base damage, but no scaling. Keep crossbow damage where it was.

Keep perception scaling on the bow.

Real-MAGNUM:
Discussed internally amongst staff, including head dev. We're reopening a new PR that should balance crossbows in a way heads agree on.

PR #818, opened 2024-05-14
[FEATURE] Real Taxes

Gyrandola:
## About The Pull Request

Have you ever wondered what exactly is taxed by the king? me too. 
A deposit fee equal to the tax percentage is applied on each money deposit in the SHYLOCK
The taxed amount is logged, together with the transaction, on the Steward's Nerve Master.

Screenshot 2024-05-15 142934

## Why It's Good For The Game

- Provides the crown with an actual stable source of income.
- Gives the king a realistic reason to even allow shylocks to exist, which is secure money storage in exchange for that money to be taxed.
- Makes taxes actually have an effect on the round as a whole. The people can rebel over high taxes.

### DIFF
--- a/code/controllers/subsystem/rogue/treasury.dm
+++ b/code/controllers/subsystem/rogue/treasury.dm
@@ -201,3 +201,23 @@ SUBSYSTEM_DEF(treasury)
 /datum/controller/subsystem/treasury/proc/log_to_steward(log)
 	log_entries += log
 	return
+
+///Calculates and deducts taxes from bank accounts, adding the amount to the treasury
+/datum/controller/subsystem/treasury/proc/tax_accounts()
+	var/total_taxed = 0
+	for(var/account in bank_accounts)
+		if(bank_accounts[account] > 0)
+			var/to_be_taxed = TRUE
+
+			//Exclude nobility from taxes. UGLY CODE
+			for(var/mob/living/character in GLOB.player_list)
+				if(character == account && character.job in GLOB.noble_positions)
+					to_be_taxed = FALSE
+					break
+
+			if(to_be_taxed)
+				var/taxed_amount = round(bank_accounts[account] * tax_value)
+				bank_accounts[account] -= taxed_amount
+				SStreasury.treasury_value += taxed_amount
+				total_taxed += taxed_amount
+	log_to_steward("+[total_taxed] taxed from the populace")
--- a/code/controllers/subsystem/rogue/treasury.dm
+++ b/code/controllers/subsystem/rogue/treasury.dm
@@ -155,29 +155,31 @@ SUBSYSTEM_DEF(treasury)
 
     return TRUE
 
+///Deposits money into a character's bank account. Taxes are deducted from the deposit and added to the treasury.
+///@param amt: The amount of money to deposit.
+///@param character: The character for whom the money account is generated.
+///@return TRUE if the money was successfully deposited, FALSE otherwise.
+/datum/controller/subsystem/treasury/proc/generate_money_account(amt, mob/living/carbon/human/character)
+	if(!amt)
+		return FALSE
+	if(!character)
+		return FALSE
+	var/taxed_amount = 0
+	var/original_amt = amt
+	if(character in bank_accounts)
+		if(character.job in GLOB.noble_positions)
+			bank_accounts[character] += amt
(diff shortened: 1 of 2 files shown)


BlueMemesauce:
idk people could just take their coins out right before the day ends 
maybe it should be taxed on transaction like putting them in or taking them out

code/controllers/subsystem/rogue/treasury.dm
```diff
@@ -201,3 +201,23 @@ SUBSYSTEM_DEF(treasury)
...
+///Calculates and deducts taxes from bank accounts, adding the amount to the treasury
+/datum/controller/subsystem/treasury/proc/tax_accounts()
+	var/total_taxed = 0
+	for(var/account in bank_accounts)
+		if(bank_accounts[account] > 0)
+			var/to_be_taxed = TRUE
+
+			//Exclude nobility from taxes. UGLY CODE
+			for(var/mob/living/character in GLOB.player_list)
+				if(character == account && character.job in GLOB.noble_positions)
```
BlueMemesauce, line 214: might break if a player mob doesnt have a job

BlueMemesauce: if character == account
cant we just do account.job?

YosemiteYam:
> idk people could just take their coins out right before the day ends maybe it should be taxed on transaction like putting them in or taking them out

This honestly. 

Do taxes currently work for the submission hole, like does any of that money go to the treasury? Because that should be the main way to collect taxes along with stuff like mammon account percentages and the merchant sales taxes

Wintermote:
> idk people could just take their coins out right before the day ends maybe it should be taxed on transaction like putting them in or taking them out

I think that's better than a once a day tax

TheManWithNoHands:
Wait... I thought taxes already apply to the shylock.
A proper tax system that applies to all citizens would be very good.
The king was already fucked over by the money spawn increase to "give the king less power" when his main power is his money.
I will die on that hill, pouring more free money into the economy isn't helping it. Fuck the starting money buff.

But as others said having the tax apply on transactions might be better than per day.
If your money slowly disappears in the shylock that kinda sucks and makes you not want to store large amounts of money in it. And the king would want everyone to store their money in the shylock because he can use the money that's in the vault.
That's why banks don't decrease your money if you store it with them, but increase it. 

Honestly having transaction tax and ALSO interest rates might be perfect. Interest rates motivate people to store their money in the vault so the king has funds he can use, and transaction tax for the king to make money. There is no reason to not look at how it works irl and copy it. If it works it works, no need to try to reinvent the wheel.

CVSthePharmacy:
> Wait... I thought taxes already apply to the shylock. A proper tax system that applies to all citizens would be very good. The king was already fucked over by the money spawn increase to "give the king less power" when his main power is his money. I will die on that hill, pouring more free money into the economy isn't helping it. Fuck the starting money buff.
> 
> But as others said having the tax apply on transactions might be better than per day. If your money slowly disappears in the shylock that kinda sucks and makes you not want to store large amounts of money in it. And the king would want everyone to store their money in the shylock because he can use the money that's in the vault. That's why banks don't decrease your money if you store it with them, but increase it.
> 
> Honestly having transaction tax and ALSO interest rates might be perfect. Interest rates motivate people to store their money in the vault so the king has funds he can use, and transaction tax for the king to make money. There is no reason to not look at how it works irl and copy it. If it works it works, no need to try to reinvent the wheel.

then just DIE on the hill bro, shut up, honestly annoying

gnomedeplume:
One idea I had was to replace the stockpile withdraw versus deposit price and replace it with a sales tax. The only problem is the current difference in withdraw vs deposit price is around 40-50% which is much, much higher than the current default tax rate of 11%. 

Taxing the shylock is probably a better solution here, just putting this idea out there in case people like it.

gnomedeplume:
This would be a very different direction but a different idea would be to have a flat tax per day instead of a percentage. Anyone who fails to pay would be marked on a ledger and the guards could go rough them up. I think this would be better than a percent per day or a transaction fee because charging a percentage discouraging people from using the shylock whereas a flat daily fee you can be punished for missing would actively encourage people to use the shylock.

Real-MAGNUM:
Testmerging, but I'm open to changes.

PR #827, opened 2024-05-15
Updates say.dm regex to cover more OOC symbols.

efzapa:
## About The Pull Request
The regex now detects any occurrence of the characters (, ), [, ], <, >, {, or } said by the player and flags it as LOOC.

## Why It's Good For The Game

OOC bad, old regex could be exploited.


BlueMemesauce:
what if people say //

PR #833, opened 2024-05-15
[ADMIN TOOLS] Aheal and Revive

Gyrandola:
## About The Pull Request

It turns out ahealing, despite it's erroneous description, never even tried to resuscitate the target. I've made a new, separate, verb that actually revives the selected mob.

image
As you can see, I grew a second head after decapitation.


Real-MAGNUM:
This shit sounds janky but I'm testmerging it anyways.

PR #840, opened 2024-05-15
Goblintrait

bobjension:
starts the inital antag conversion for gobbos by adding a proper trait so they can only access their cave

DOES NOT ADD IT TO ACTUAL GAMEPLAY


Real-MAGNUM:
Conversion into antag is still in discussion.

ChungusGamer666:
conflicts

PR #845, opened 2024-05-16
Necrodus

Stagrawl:
Kicks out the Necralytes from the church because no one likes them and they're weird. 

Gravediggers have taken up the mantle due to their time spent in the graveyards performing the burial rites of the Undermaiden and a familiarity with death.

This will be a good thing, I promise.

### DIFF
--- a/code/modules/jobs/job_types/roguetown/peasants/gravedigger.dm
+++ b/code/modules/jobs/job_types/roguetown/peasants/gravedigger.dm
@@ -1,44 +1,42 @@
 /datum/job/roguetown/undertaker
 	title = "Gravedigger"
 	flag = GRAVEDIGGER
-	department_flag = PEASANTS
+	department_flag = CHURCHMEN
 	faction = "Station"
 	total_positions = 2
 	spawn_positions = 3
+	allowed_patrons = list(/datum/patron/divine/necra)
 
 	allowed_races = list(
 		"Humen",
 		"Elf",
 		"Half-Elf",
 		"Dwarf",
+		"Aasimar",		
 		"Tiefling",
 		"Argonian",
 		"Dark Elf",
 		"Aasimar",
 		"Half Orc"
 	)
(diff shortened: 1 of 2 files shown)


Real-MAGNUM:
Highly controversial. Likely will be closed. If they're removed, it'll be done after internal discussion. @ChungusGamer666

Stagrawl:
I was asked to do this! :)

The reasoning here goes something along the lines of Necralytes and Gravediggers sort of taking up the same space in dealing with the dead. There's a bit of toe-stepping in that regard.

Shifting Necra Devotion over to Diggers means that (while not likely to happen) there could be 5 Necra Acolytes running around, with even more possible if clerics rock up. This change ensures more diversity of faith and so on.

This change also gives Gravediggers reign over the realm of the dead, ensures the church has access to Necra's utilities and forms a slightly closer bond between the graveyard which observes the sacred burial rites of one of the Nine and the church itself.

Real-MAGNUM:
Please reference who asked you to make a pr if it was suggested by a maintainer or head of staff

Stagrawl:
I hesitate to outright say "Abraxas said to do it" but this was one of the conditionals of making Gravediggers Necralytes.

image

Real-MAGNUM:
You're missing a necra hood in their loadout. Make sure you add them on gravediggers. Add that and you have my support.

The same thing will eventually happen to noc acolytes too. They'll be merged into the court wizard stuff in one way or another ONLY once spellscribing is complete.

Stagrawl:
I think Nocalytes were possibly (also?) going to be shoved into the Archivist from discussion.

code/modules/jobs/job_types/roguetown/peasants/gravedigger.dm
```diff
@@ -1,52 +1,57 @@
...
 	total_positions = 2
 	spawn_positions = 3
+	allowed_patrons = list(/datum/patron/divine/necra)
 
 	allowed_races = list(
 		"Humen",
 		"Elf",
 		"Half-Elf",
 		"Dwarf",
+		"Aasimar",
```
ChungusGamer666, line 15: redundant, was already in the list

ChungusGamer666:
Will take over and handle it myself

PR #846, opened 2024-05-16
RACISM LORE V4

ChungusGamer666:
## About The Pull Request

No regerts
-Removes half orcs from templar
-Removes all non-humen except half elf from desert riders
-Removes all non-humen except half elf from grenzelhoft mercs


CVSthePharmacy:
why did you remove tiefling from desert riders

PR #855, opened 2024-05-16
recurve bow, iron messer, and iron short sword

cornercube:
## About The Pull Request
adds the recurve bow
sprite thanks to "infinx" on discord
deals slightly less damage, but costs less stamina to hold fire and charges to full power faster
bows now have their own var, damfactor, that multiplies the damage of their arrow
so now we can have different bows that deal different damage without needing new arrows
bow has 1x damfactor, recurve bow has 0.9
also adds a recipe for the standard bow
crossbow bolts now have a damage of 35, while the crossbow itself has a damfactor of 2
damage is the same, but the bolt is no longer a doom knife

adds the iron messer
a single edged, single-handed, iron sword
sprite thanks to "mortiz" on discord
large knife might be more descriptive
it has a fast chop attack, a generic thrust attack,  and very poor defensive stats

iron short sword has it's sprite fixed slightly
it should be even shorter and a more visible handguard
removes hanging pixel to conform to spriting standards

old iron sword sprite re-added, iconname is now ironswordalt
in case anyone wants a use for it
maybe a slashing weapon

iron short sword also has irs STR requirement decreased to 4, making it a true universal weapon even a prince can use
iron sword had 6 STR req, but when wielded it goes down to 3
inability to be wielded means that the short sword could actually be harder to use

ghe

## Why It's Good For The Game
iron superiority
recurve bows give slightly more ranged variety
craft able bows are a long time suggestion
crossbows might be easier to balance if the bolt they fire isnt so powerful to begin with

bow recipe:
shameless copypasta of stake code on a small log: use knife on log to get an unstrung bow
craft a bow string from 2 fiber
craft a bow from an unstrung bow and a bow string (craftdiff = 2)


Takyon69:
Can you please remove the 'tip' on the short sword.

TheManWithNoHands:
Please tell me you also add a way to craft the new bow.

Real-MAGNUM:
I support both comments above.

cornercube:
> Can you please remove the 'tip' on the short sword.

barp
like this?

> Please tell me you also add a way to craft the new bow.
was planning on making it unique to rangers and hunters and could only be ordered by merchant for now
not until we have glue, horn, and sinew can we really make it craftable

we also cannot craft regular bows in any case
I could add a placeholder recipe for both the bow and the recurve bow, though

TheManWithNoHands:
> we also cannot craft regular bows in any case I could add a placeholder recipe for both the bow and the recurve bow, though

Please do. The previous Pr that tryd to add a bow recipy got abandond and closed. Bows would be in such a better spot if you could actually get your hands on em.

cornercube:
> Please do. The previous Pr that tryd to add a bow recipy got abandond and closed. Bows would be in such a better spot if you could actually get your hands on em.

added the ability to carve logs into unstrung bows, the ability to craft bowstrings from fiber, and make a wooden bow

code/game/objects/items/rogueweapons/ranged/ammo.dm
```diff
@@ -18,7 +18,7 @@
 */
 /obj/projectile/bullet/reusable/bolt
 	name = "bolt"
-	damage = 70
+	damage = 35
```
JTGSZ, line 21: ?????

code/game/objects/items/rogueweapons/ranged/bows.dm
```diff
@@ -19,6 +19,7 @@
 	cartridge_wording = "arrow"
 	load_sound = 'sound/foley/nockarrow.ogg'
 	associated_skill = /datum/skill/combat/bows
+	var/ damfactor = 1
```
JTGSZ, line 22: huh

code/game/objects/items/rogueweapons/ranged/bows.dm
```diff
@@ -67,7 +68,7 @@
 		else
 			BB.damage = BB.damage
 			BB.embedchance = 100
-		BB.damage = BB.damage * (user.STAPER / 10)
+		BB.damage = BB.damage * (user.STAPER / 10) * damfactor
```
JTGSZ, line 71: You know theoretically if persay... lets use 11 perception on the user in these scenarios
We have 30  * (11 / 10) it'd come out to 33 on default
With your damfactor on the recurve bow it'd be 
30 * (11 / 10) * 0.9 which would come out to 29.7
Even if it was a intended damage reduction, you should prob round() it.

code/game/objects/items/rogueweapons/ranged/ammo.dm
JTGSZ: this makes more sense once I realize your damagefactor multiplies things by two on the crossbow.

PR #871, opened 2024-05-16
Light armor changes

LucaNotLucas:
## About The Pull Request

**Light armor dodging**
Adds a trait called "Fast reflexes" for rogues, rangers, and generally fast characters, makes it so dodging is easier and takes less stamina while only wearing light armor.

**New armor and crafting**
Adds a new light armor (since there isn't that many other than gamberson and leather), heavy leather armor! Takes 2 furs and 1 hide to make, protects arms, chest, groin and vitals, has some decent durability, and looks perfect on a rogue. Also added the ability to craft heavy leather gloves for 1 fur.

## Why It's Good For The Game

First, adds more variety when it comes to light armors, gives more uses for furs and hides, gives better light armors, and makes dodging actually an option, making it *maybe* easier to dodge than to parry, at the cost of not being able to wear medium/heavy armors.

### DIFF
--- a/code/modules/roguetown/roguejobs/alchemist/reagents.dm
+++ b/code/modules/roguetown/roguejobs/alchemist/reagents.dm
@@ -38,6 +38,7 @@
 	description = "f"
 	reagent_state = LIQUID
 	color = "#00B4FF"
+	taste_description = "bitter"
 	metabolization_rate = 0.1
 
 /datum/reagent/berrypoison/on_mob_life(mob/living/carbon/M)


Wintermote:
Doesn't feel like a good idea to tie armor usefulness to traits, rather than making it intrinsic to the armor as a balance factor; I'd personally do away with the heavy/medium training entirely for example and bake things like dodge/speed adjustments into the armor itself.

OrbitalParadise:
I agree that taking off medium armor shouldn't take away traits from your character, I think it would be better to allow the trait to work with medium armor/no armor and then if the player decides to put on heavy armor then the trait dissapears until it is taken off.

LucaNotLucas:
> Taking off medium armor shouldn't take away traits from your character, I think it would be better to allow the trait to work with medium armor/no armor and then if the player decides to put on heavy armor then the trait dissapears until it is taken off. Not sure if that is how the code works here.

The code checks if the person has medium/heavy armor, and if the person has the trait, if the person has armor or doesn't have the trait, returns false and then it gives the normal stats for dodging, if the person doesn't have medium/heavy armor and has the trait, it returns true, wich gives higher chance to dodge and less stamina drain.
The traits are never changed, it just checks if you have it or not, and applies the buffs if you do.
Also, medium armors already protect from 100% of damage, some integrity and have their own trait, so making it also so you can dodge easly with them would just make them objectively the best armor.

> Doesn't feel like a good idea to tie armor usefulness to traits, rather than making it intrinsic to the armor as a balance factor; I'd personally do away with the heavy/medium training entirely for example and bake things like dodge/speed adjustments into the armor itself.

While yes, i agree with you, this would mean that most beggars/regular towners/etc. would have a really high chance to dodge, wich would make them as skilled in fighting (defensively) as somone with actual training at dodging, not only that but heavy and medium armors already have traits to be able to decide who can wear them effectively, so not everyone can just wear the best armors, and get the same benefits as everyone else.

PR #876, opened 2024-05-17
Quality Control MK1

JTGSZ:
Formerly dead, I've been summoned to do some quality control via my discord dms.
This PR does the following:
Disables Goblin jobs/species for the moment - They are half-baked, loaded with bugs and other things that need to be addressed before they are in live play. Secondly, there are friendly goblins in the town.

Thus for goblins to go back live:
They need polished until emoney is happy with the state of them.
They need to be made antags (probably swap the 20 bandits out sometimes with 20 goblins + a kingdom for some spicy variation)
(The fact there were/still are AI goblins should have still clued you in on this as I stated before my departure)

Half-Orcs removed from most garrison jobs outside of Bog guard
Rockhill is a human town in lore, and they are a bit too inhuman looking despite being half-human/more human than some of the other things in there. (haha, small details not that you'd know much about those!)

Warning Removals:
The person who mapped in the blackstone map also left their map file ticked in instead of using the standard loading file, not that it matters but its still going to load the map anyways, thus blackstone.dm was unticked, and the rt.dm map load is used.

The person who decided to remove the (actually really shitty) perception modifier on some projectiles removed the perception conditional check that set the shit higher, but ignored the loop, resulting in a warning that it was useless. That was also removed.

### DIFF
--- a/code/game/objects/items/rogueweapons/ranged/crossbows.dm
+++ b/code/game/objects/items/rogueweapons/ranged/crossbows.dm
@@ -106,8 +106,6 @@
 			spread = 150 - (150 * (user.client.chargedprog / 100))
 	else
 		spread = 0
-	for(var/obj/item/ammo_casing/CB in get_ammo_list(FALSE, TRUE))
-		var/obj/projectile/BB = CB.BB
 	cocked = FALSE
 	..()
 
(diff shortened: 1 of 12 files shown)


CVSthePharmacy:
it's all panvx's fault kill him

PR #877, opened 2024-05-17
Attempts to stop noble roles from disappearing thanks to being picked for vampire

JTGSZ:
Now that I'm not in the grave for the moment, time to attempt to solve more random crap brought up.

The king was was in the restricted jobs list but somehow still ended up being converted into a vampire thus removing the king from the round.

I skimmed slightly but didn't feel like checking everything for how exactly a king is becoming a vampire on round start from a random anecdote and also found nothing notable while skimming.

Thus I at best can just cram in extra checks on both vampire(and werewolf) in an attempt to stop nobles from being sucked in past the restricted jobs list part.


JTGSZ:
I looked at things again and I also kinda looked at the round it occurred in and I have a better clue on what happened, and it might not be vampire related despite the player saying so. So gonna close this and add some more checks in the gamemode's antag selector.

PR #883, opened 2024-05-17
BarbarianGODS - Swaps medium armor training for critical resistance

ChungusGamer666:
barbarians cannot be instakilled by crits instead of wearing armor like PUSSIES


zakanater19:
the new meta class lol

PR #892, opened 2024-05-17
delf desert riders

TopHatPenguin:
With the likelihood that the DR mercs will be based out of a delf citystate, it would make sense for the delfs to actually play the role, especially as it fits into the outsider parameters that delfs originally had.

This is a tiny change pr but for better standards here it is tested and working:
https://gyazo.com/7c929fb89ba12aaa486d7ed1de7b4028


ChungusGamer666:
Made without my permission.

TopHatPenguin:
Under review by Abraxas and the lore team until further notice.

PR #909, opened 2024-05-18
Simple mobs parrying fix

Bakeun0:
## About The Pull Request

Null values screwing everything up, parrying will now work against simple mobs.

This is just a small fix before I finish completely re-writing the proc, there are still a few bugs such as not using a valid parrying tool if your active hand is on a non-parrying item, and fist parries not working properly against weapons.

## Why It's Good For The Game
 
Minor bug fix


ChungusGamer666:
wow this proc is bad

PR #910, opened 2024-05-18
Bows borked for some classes, pls fix

Stagrawl:
## About The Pull Request

Makes bows spawn, hopefully.

## Why It's Good For The Game

The bows will spawn for Hunter + Hunter Killer.


JTGSZ:
Weird, the hunter killer has a override on that slot with a knife, but the hunter has nothing blocking it from going into the left back slot.

PR #912, opened 2024-05-18
LOOTBOXES

Alibouk1:
FOR LOOTIN'


Real-MAGNUM:
idk what's with the .dme edits

Alibouk1:
Oh idk why those happened either, you should be able to modify it from here though I think ?

roguetown.dme
```diff
@@ -8,6 +8,8 @@
 // END_PREFERENCES
 // BEGIN_INCLUDE
 #include "_maps\_basemap.dm"
+#include "_maps\blackstone.dm"
```
JTGSZ, line 11: It is important not to include all the files that try to redefine what map is loaded bro....
Also someone commented out necromander in includes so technically it is fine to remove that, but still lol

Gyrandola:
Screenshot? Where do they spawn? What do they look like?

roguetown.dme
Alibouk1: It should be fixed now ? You can double check, I might've misclicked something earlier otherwise I'm not sure why this file changed at all.

Alibouk1:
> Screenshot? Where do they spawn? What do they look like?

I used the haunt bonepile as a placeholder for now, they're not in the map yet but should be easy enough to add as a loot drop for monsters later on.

roguetown.dme
JTGSZ: Correct, just your lootpiles and one automatic movement of the hand of heartfelt on the list i suppose

PR #916, opened 2024-05-19
Checkdefense re-write

Bakeun0:
## About The Pull Request

Went over the entire checkdefense proc and pretty much removed and changed anything that I thought was unnecessary or a duplication, then sorted it in a way that doesn't make my eyes bleed.

This also includes a few changes and bug fixes:

- You will now ALWAYS use the item with the highest chance to parry, no matter what hand is currently active.
- Unarmed parrying chance is increased by 20% per skill level like every other weapon.
- You will do an unarmed parry by default if your unarmed parry chance is higher than the currently held weapon.
- Skill gain will be lower than usual (if it's supposed to be higher just increase by the full amount of int, instead of half).
- Being on help intent no longer forces you to fail an unarmed parry.
- Removes penalty for "improvised" weapons (shields will need to have their defence reduced by 1 if you want their parry chance to stay the same).

Still needs to be fixed:

- Unarmed parries don't work against weapons if you aren't holding anything. It will say the parry was successful but you still take damage, although this time it's not the result of null value shenanigans. For some reason it works properly if you're holding any item in either hand.

## Why It's Good For The Game

ChungusGamer666 commented: wow this proc is bad

### DIFF
--- a/code/modules/mob/living/roguetownprocs.dm
+++ b/code/modules/mob/living/roguetownprocs.dm
@@ -266,54 +266,6 @@
 					testing("failparry")
 					visible_message("<span class='info'>fail2</span>")
 					return FALSE
-/*
-				testing("begin unarmed parry")
-				if(!has_hand_for_held_index(active_hand_index))
-					testing("noarm unarmed parry")
-					return FALSE
-				if(H)
-					if(H.check_arm_grabbed())
-						testing("grabbed unarmed parry")
-						return FALSE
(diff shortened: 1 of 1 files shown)


Bakeun0:
Could keep working on this to make it better but for now I want to start some of the wrestling/short blades synergy stuff I had planned a while ago.

ChungusGamer666:
looking good!

PR #921, opened 2024-05-19
Properly poisons the well: Organ poison and berry poison have no taste, they are unnoticeable

ChungusGamer666:
title


dunsel-hordon:
So what you are saying is you just implemented in a useless bloat system that simply  makes us take a couple seconds longer to identify if something is poison or not because if you take one bite you are fine.

CVSthePharmacy:
this PR sucks

PR #922, opened 2024-05-19
Trimming the herbs - Removes necra acolytes, renames gravediggers to morticians and turns them into a church role

ChungusGamer666:
## About The Pull Request

Title

## Why It's Good For The Game

Currently, necra acolytes serve no niche that the gravedigger does not - In fact, they constantly get in the way with their spells. This is just a chance to ensure there is no such conflict of interest and that no acolyte slots are being wasted on what is essentially just gravedigger+.


Stagrawl:
Yeesssss

PR #928, opened 2024-05-19
Eat your eye out - Replaces most instakill crits on the head with being crippled for life

ChungusGamer666:
## About The Pull Request

Eyestab will now poke out the eye you're aiming for, permanently altering the FOV of the affected. If both eyes are poked out, the eye gets gored, and thus removed from the owner. Eyestabs can still brainkill both only once the eyes get gored!
Tongue stab will now tear off the tongue instead, and will never instakill.
Nose stabs do nothing because I couldn't figure out a fun effect for them, maybe disfigurement could be done later?

## Why It's Good For The Game

Fuck instakills, being crippled is better.


DirtTastesBad:
Could have nose... increase stamina use? Harder to breath I'd imagine

efzapa:
throat cut spambros we are so back

Hmeister-real:
a little late, but just removing noses would be good, even just as a sprite change so that i can remove criminals noses and have them permenantly labelled as "criminal" without disabling them

PR #931, opened 2024-05-19
[FIX] Grenzelhoft accent less intruding "nein"

Gyrandola:
## About The Pull Request

Nein should no longer be translated within larger words, at the price of not being able to appear unless there is a word after it.


zoktiik:
THANK YOU
THIS IS THE SINGLE BIGGEST PAIN POINT WITH GRENZELHOFT ACCENT
I HATE "NEINW" SO MUCH IT'S UNREAL

PR #935, opened 2024-05-19
You can now find gems from mining

lovegreenstuff:
## About The Pull Request

See title.
Mining can now drop gems and very very rarely a riddle of steel. 
Adds new gem rings that we already had sprites for.
Also rebalances gem ring sale prices because it makes no sense to lose money doing this. Thankfully, no one spawns with gem rings so this won't really change the economy all that much.

## Why It's Good For The Game

i dunno lol

### DIFF
--- a/code/game/objects/items/rogueitems/gems.dm
+++ b/code/game/objects/items/rogueitems/gems.dm
@@ -29,7 +29,7 @@
 /obj/item/roguegem/green
 	name = "gemerald"
 	icon_state = "emerald_cut"
-	sellprice = 22
+	sellprice = 42
 	desc = "Glints with verdant brilliance."
 
 /obj/item/roguegem/blue
@@ -41,7 +41,7 @@
 /obj/item/roguegem/yellow
 	name = "toper"
 	icon_state = "topaz_cut"
-	sellprice = 14
+	sellprice = 34
 	desc = "Its amber hues remind you of the sunset."
 
 /obj/item/roguegem/violet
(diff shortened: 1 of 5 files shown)


lovegreenstuff:
oh wait we have sprites for rings that we can't craft give me a sec

ophaq:
Finally, some good shit. Been wanting gems. It made no sense for all these rocks to contain gold and ore near lava and no gems. Riddle is a nice touch too as long as its not too rare but not too easy either.

lovegreenstuff:
diamond i can use the sprite for but i need a name for it since all the gems have unique names
give me like 20 minutes

Real-MAGNUM:
Looks balanced in terms of mapgen

Knightfall5:
>Riddle is a nice touch too as long as its not too rare but not too easy either.
> 


Assuming you are mining in the best area, each tile has a .04% chance of giving you a riddle of steel. So, on average, one every 2000~ tiles mined. Good luck!

dunsel-hordon:
Wait can we actually mine silver now?

ophaq:
> > Riddle is a nice touch too as long as its not too rare but not too easy either.
> 
> Assuming you are mining in the best area, each tile has a .04% chance of giving you a riddle of steel. So, on average, one every 2000~ tiles mined. Good luck!

I think it should be like, 0.5% or 0.1% on the most valuable ore tile. Riddles are valuable but aren't that valuable. Since mining takes forever anyway and you mine way less than around 100ish tiles considering you're solo, or maybe a lot more (but still under 2000) for an entire team of people. At that rarity, you are just better off selling gold rings and buying from the merchant.

PR #948, opened 2024-05-20
Brings back not being able to play the same job twice in a row

ChungusGamer666:
## About The Pull Request

Awful PR merged while I wasn't head dev. I never agreed with it, now I can rectify it.

## Why It's Good For The Game

Players should generally strive to be varied rather than fall on the same dumb routines, this ensures that only overflow roles can be played over and over again (beggar, pilgrim, towner, lunatic)


BlueMemesauce:
merchant ahould bypass last class because its a required role
why should hand bypass last class?

PR #953, opened 2024-05-20
Update elfd.dm

Ambrose997:
Slightly changes the stats for male dark elves so the debuffs are less harsh. INT is now 2, Endurance and Constitution are both -1, Strength increased slightly to be more on-par with the strength debuffs between male and female in other races like dwarves, half elf etc. 

## Why It's Good For The Game

Male drow are dis-proportionally debuffed when you consider that similar debuffs (i.e men being stronger than women) do not usually have this much of a disparity and it's a bit inconsistent. Aside from the case with human (where women have more INT than men), dwarf, half elf, and wood elf all have the same int for both male and female. 

As a result this has led to barely any male drow being seen in-game due to it simply just not being worth it at all leading us to only seeing female drow all the time (cringe). Femstatics on other races get played more frequently despite being weaker than the male variant, so it's somewhat proven that you can still have a weaker sex that people will still play as long as the nerfs aren't disproportionate. 

This change allows for the disparity to still exist whilst not being harsh to the point where it's considered non-viable, which hopefully means more people will play it and provide some variety to the race. Before anyone bitches: Beggars recently just got a buff to master-legendary stealing skill for the same reason, so I do not think this is too much to ask for.

and lastly: merge this or gay


Gyrandola:
image

Ambrose997:
> image

I literally asked him beforehand if this would be okay.

Stagrawl:
Dark Elf males should remain submissive. 

Double the debuffs, make them more pathetic, then give them female cooking skills.

Stagrawl:
So long as they remain worse than women, it has my support ig.

CVSthePharmacy:
just remove drows and dark elves, I don't like them

PR #956, opened 2024-05-20
Area fixes + new area intro texts

Vivalas:
## About The Pull Request
Fixes area Entered proc being split in two for some reason. This overrode the first proc and ignored a vital `set waitfor= FALSE` line that caused pulling to lag across area boundaries.

Also, since area titles work now, added one for the town and the keep. This may also enable some mood / beauty things that I'm not sure how exactly they work and might cause some balance issues, so this potentially warrants a testmerge before a full merge.

## Why It's Good For The Game
Fixes something broke.

### DIFF
--- a/code/game/area/areas.dm
+++ b/code/game/area/areas.dm
@@ -611,7 +611,7 @@ GLOBAL_LIST_EMPTY(teleportlocs)
   *
   * If the area has ambience, then it plays some ambience music to the ambience channel
   */
-/area/Entered(atom/movable/M, area/old_area)
+/area/Entered(atom/movable/M, atom/OldLoc)
 	set waitfor = FALSE
 	SEND_SIGNAL(src, COMSIG_AREA_ENTERED, M)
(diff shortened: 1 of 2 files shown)


code/game/area/areas.dm
```diff
@@ -611,7 +611,7 @@ GLOBAL_LIST_EMPTY(teleportlocs)
   *
   * If the area has ambience, then it plays some ambience music to the ambience channel
   */
-/area/Entered(atom/movable/M, OldLoc)
+/area/Entered(atom/movable/M, area/old_area)
```
JTGSZ, line 614: You probably want to keep that param as OldLoc and not area/old_area

ChungusGamer666: yeah that is no bueno

Vivalas:
Was going to ask what the actual issue was there but then just checked the DM reference instead.

JTGSZ:
Correct, you typed it up to area/old_area from atom/OldLoc, most of whats being passed into that param is a loc which is a reference to a turf afaik.

Vivalas:
Yes I don't know why the current implementation uses area.

PR #972, opened 2024-05-21
Chests, ores and bowl resprited

Takyon69:
## About The Pull Request

Chest resprited with a gold and steel version. Needs coding to replace the current one. Ores have been resprited and now have 3 variations which also need to be coded. Wooden bowl resprited for cooking and doesnt need to be coded :)

image

image

image

## Why It's Good For The Game

Aesthetics.


ophaq:
Looks good. Could we perhaps get wooden plates too?

CVSthePharmacy:
I think the old chest looked better

TheManWithNoHands:
You plan to change the function of the chests or why do you need coding for that? Looks like a simple resprite after all.

Takyon69:
> You plan to change the function of the chests or why do you need coding for that? Looks like a simple resprite after all. Oh i see. This also adds a second chest version without the middle bar.

Basically just keeping the chests we have incase we need them for something, though that's unlikely. Just want the default to be random. So there would be both gold/silver chests for variety.

TheManWithNoHands:
> Basically just keeping the chests we have incase we need them for something, though that's unlikely. Just want the default to be random. So there would be both gold/silver chests for variety.

Would have replaced normal chests with them. But alrigth. Want me to just commit the code for the chests? Dont know how to contribute to PR'S at all.

Takyon69:
> Ok, random i cannot do. Also i wont do that, because thats a bad idea. On so many levels. That just fucks over any mapper that wants to use specific chests for any stetic. I get that you want the new sprites used, but people are gonna replace the old chests with the new ones over time where it fits.

Yeah, fair enough about the mappers. Not something I initially thought about. Either way I want to replace the old chest sprites because they suck.

TheManWithNoHands:
> > Ok, random i cannot do. Also i wont do that, because thats a bad idea. On so many levels. That just fucks over any mapper that wants to use specific chests for any stetic. I get that you want the new sprites used, but people are gonna replace the old chests with the new ones over time where it fits.
> 
> Yeah, fair enough about the mappers. Not something I initially thought about. Either way I want to replace the old chest sprites because they suck. Whether they're randomly steel or gold upon map generation doesn't really matter. If it makes things easier for mappers, what we could do is simply replace the chest we have now with just the steel version and later on I can make various different chest themes which mappers can play with.

That would be very nice, but the diffrent chests you already made are pretty dope.

PR #977, opened 2024-05-22
ghetto farming

lovegreenstuff:
## About The Pull Request

Adds an inbetween for no tools and max tools farming, ghetto farming
Wooden clubs can be used to thresh grain, just not all of it at a time
Anything capable of cutting in the offhand (note, very important) while harvesting will reduce the time and increase harvest chance for the unskilled
A shovel or spade in the offhand (still very important) will decrease time to weed by half.

## Why It's Good For The Game

Playing towner farmer is painful, and so is playing soilson when all your tools have been stolen. This alleviates that somewhat.

### DIFF
--- a/code/modules/roguetown/roguejobs/farmer/chaff.dm
+++ b/code/modules/roguetown/roguejobs/farmer/chaff.dm
@@ -26,7 +26,7 @@
 		new /obj/item/natural/fibers(loc)
 		qdel(src)
 
-/obj/item/natural/chaff/attackby(obj/item/I, mob/user, params)
+/obj/item/natural/chaff/attackby(obj/item/I, mob/living/user, params)
 	testing("attackb")
 	if(istype(I, /obj/item/rogueweapon/pitchfork))
 		if(user.used_intent.type == DUMP_INTENT)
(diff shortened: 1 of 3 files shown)


ophaq:
Good PR, having a knife or something in off hand would be helpful instead of needing a sickle all the time. Same with the club thresher. Puts clubs to more use too.

YosemiteYam:
Makes farming abit too easy with little interaction with smiths to make sickles/buy it from merchants. If this is merged then using a cutting weapon that isnt a sickle should damage the weapon abit each time, to dissuade someone from using this as a long term solution.

lovegreenstuff:
> Makes farming abit too easy with little interaction with smiths to make sickles/buy it from merchants. If this is merged then using a cutting weapon that isnt a sickle should damage the weapon abit each time, to dissuade someone from using this as a long term solution.

i mean like sure its not terribly difficult to do that but like stone knives are easy as shit to make so like

ophaq:
> Makes farming abit too easy with little interaction with smiths to make sickles/buy it from merchants. If this is merged then using a cutting weapon that isnt a sickle should damage the weapon abit each time, to dissuade someone from using this as a long term solution.

Smiths are so busy currently that half the time I spent waiting half the round to buy one thing. The less work they have to do the better.

PR #983, opened 2024-05-22
New Court Physicans room, update to keep armory and house keeping

TheManWithNoHands:
## About The Pull Request
* Removes the keeps weird upper library and replaces it with the court physician's own room. Containing medical supplies, workstations, beds, and a balcony to grow medical herbs. Also gives them the magican area because the music is nice.
* Expands the Keeps armory to fit the new Keeps guard spawn gear. Mostly Exchanging gear with its steel equivilant, adding bows and arrows, some spare keys and in general more gear. The keeps armory should be more impressive then the Bog's.
* Housekeeping, fixing map errors.

*You hear a Zad in the distance*

## Why It's Good For The Game
https://imgur.com/a/vVTvKCC

The court phys is a Nobel employed by the king to keep him and his family alive. They kinda need their own room. This also preps for when the alchemy system is properly implemented and the court phys can make medicine.

Keeps armory should have gear in line with the keeps guard spawn gear, and it should be better than the bog armory. Mainly added some more space and switched iron gear with steel. If the gear present is too much, that's easily fixed so no trouble there.

Map fixes good. I found a part of the bog copyd and floating a z-lv above bro.


Sarkness:
I deny this one purely because if you remove this library, the Archivist has nowhere to be anymore.

TheManWithNoHands:
> I deny this one purely because if you remove this library, the Archivist has nowhere to be anymore.
> 
> But otherwise, must that pass. Very good work, honestly.

Not the Liberary UNDER the keep. The second one only the king has acces too, and thats so under used that you even thougth of it but the liberians library instead.

YosemiteYam:
Maybe add one or two crossbows instead of bows to the keep armory like the town watch barracks, since these guys are men at arms and would be the only people that would have a couple of crossbows laying around to defend against a siege of peasant bowmen/ranger rebels etc. 

Maybe for the kings library (thats basically his study in addition to his office) divide it into two, make one part the physicians table and one part keep the bookcases since a court physician would still be kind of a scholar that would need access to books on ailments/creatures/general knowledge etc.

TheManWithNoHands:
> Maybe for the kings library (thats basically his study in addition to his office) divide it into two, make one part the physicians table and one part keep the bookcases since a court physician would still be kind of a scholar that would need access to books on ailments/creatures/general knowledge etc.

Its already devided into two on the current build. His study with tablesa dn such has a whole side just filled with bookshelfs. All this room is, is additional bookshelfs. And the fact that there are already a couple bookshelfs in the keep, and a much larger library, makes this small private liberary (that doesnt even look nice) kinda pointless. But a bookshelf i will think about.
Also maybe on the crossbows. Will take a look.

TheManWithNoHands:
> this would be sweet!

redoing the Pr rigth now. Because darn map merge conflict. Will ping Magnum immediatly after the new Pr is up.

PR #997, opened 2024-05-22
[SPLIT] Sarkness Extravaganza: Noc Acolyte Final Spell: Moonbeam

Sarkness:
## About The Pull Request

Noc Tier 3 Spell: Moonbeam. Concentrate the moon power on someone to incapacitate them highly, or to completely annihilate any undead being too dangerous for the Church.

## Why It's Good For The Game

Largely followed and appreciated for it's D&D ressemblance, this V1 should give a preview of the ray of light that'll soon gives heavy sunburn to those unfortunate into walking into the pillar of divine justice.


CVSthePharmacy:
Back in my day these used to gib people

YosemiteYam:
We really should figure out better spells other than = "This spell is so strong it incapacitates you outright or just kills you!". Especially since you have no way to block them unless if you're lucky enough to dodge a histscan spell. 

The undead part also doesn't make sense to me, what relation does Noc have to undead? Why would mages have an undead turning spell (All noc spells are moving to mages since noc acolyte is being removed). This doesn't make sense in that context. 

More mage spells should be in the lines of small to medium damage ranged spells for offense with variety (think magic bolts) or utility/mobility/buff/debuff spells for variety. Things like slow, things like "Dehydrate" (gives you the worst third demodifier and lowers your endurance due to it), things like random teleport that teleports you in a random direction for 4-5 paces but takes a long time to charge again. 

Not another stun combat offender of a spell that also does a shitton of damage

PR #1001, opened 2024-05-23
Skill learning

ShadowLeaderJordanzo:
## About The Pull Request

adds exp gain to climbing (trees only) / pickpocketing, also lets you know if a pickpocket is impossible (if u have >= perception as target)
the latter can b removed if its too metay

## Why It's Good For The Game

makes it so more skills gain exp now, idk youth buff or something witty


ChungusGamer666:
looks ok

Ambrose997:
Pretty sure this doesn't work. I still don't gain levels in climbing no matter how much I climb.

Ambrose997:
I also weirdly can't climb stone walls anymore despite having apprentice climbing (would work before).

PR #995, opened 2024-05-22
[SPLIT] Sarkness Extravaganza: Blacksmithes Nerf

Sarkness:
## About The Pull Request

Nerfed Male Blacksmith Pilgrim, Buffed Female Blacksmith Pilgrim so they're on par together and gives a breather/importance to the Weaponsmith/Armorer
Also buffed lightly weaponsmith/armorer

## Why It's Good For The Game

A huge disparity between both blacksmithes (credits to @wow5000) which made female blacksmithes unable to forge properly with the new Smithes PR while Male Smithes Pilgrims outclassed Armorer/Weaponsmiths, making these normally crucial role very much useless. This should offset the balance and make it an equal ground for the smithes pilgrims, and the armorer/weaponsmith while keeping advantage to males without being detrimental to females smithes players.

PR #1004, opened 2024-05-23
Blacksmith changes

ShadowLeaderJordanzo:
## About The Pull Request

various blacksmith changes

- gives weaponsmith/armorer 4-5 starting skills, changes old from giving 1-2 to 1. Since masterblacksmith starts at 6 in both, let the ooak jobs get 4-5 (5-6 if old).
- makes the white strike give 2x progress, after calculations. chance is 7% per level of blacksmithing, 23 prob was the original, so at 3 skill level, you have the same chance as before.
- adds crafting difficulty, pls note the comment on the push 
(I AM NOT A BLACKSMITH OR HISTORIAN
the logic behind the diff is simple:
if it takes 1 more bar +1, if it takes 2, +2, etc.
valuables start at 2 diff, weapons and armor at 1, tools and other at 0)
- crafting diff reduces the prog you gain by 6 per tier of diff, so a legendary making a diff 4 will be essentially a novice, they still keep the 1.5x prog chance

## Why It's Good For The Game

makes blacksmithing more engaging


Sarkness:
Would recommend you add my nerf into the mix, maybe you would be able to find true balance with the changes I've made for the pilgrim blacksmith so your armorer/weaponsmith are balanced, probably even better than the blacksmithes?
[Blacksmithes Nerf](https://github.com/Blackstone-SS13/BLACKSTONE/pull/995)

wow50000:
not a fan of the rng involved in starting skills and i think buffing the pilgrim smith further might unironically kill the towner smiths on the spot. not sure about the balance about white strike giving 2x and increasing the difficulty so that will need testing

wow50000:
so probably keep the skills static at 2-3 and remove RNG would be my opinion on this

wow50000:
also mentioning i am for these changes but my only issue here is the RNG(not your fault) and the higher starting skills could completely fuck armourer and weapsmith

wow50000:
armourer and weapsmith should spawn with 4. or 5 if old
merge both statblocks for male and female smith pilgrim and remove all the RNG from both but abraxas said remove only some of it
as for craft diffs remove the impossible and give it testing idk but all smiths should be able to craft anything as the current items stand but some legendary armours could be a good idea for being only craftable by master smiths or so on

PR #1006, opened 2024-05-23
Vaquero touch-up

GreatRuna:
## About The Pull Request

changes 'if' to 'else' for vaquero cloak chance, to try and fix the bug of it only spawning red. gives vaquero one more point in swords.

## Why It's Good For The Game

The code does not work as intended, with the cloak roll, and fixing it gives that code purpose. 

Giving Vaquero 4 in swords makes sense, because that is their thing. They are swashbucklers, and a swashbuckler being better at riding than swordfighting is very odd. It is balanced by the fact they do not have armor at start and cannot wear heavy or medium armor, nor do Vaqueros start with any ranged weaponry.


GreatRuna:
I dont know why the combat music line was even touched, i didnt edit it, it just did that

code/modules/jobs/job_types/roguetown/adventurer/types/combat/donator/vaquero.dm
```diff
@@ -42,7 +42,7 @@
 	cloak = /obj/item/clothing/cloak/half/red
 	if(prob(33))
 		cloak = /obj/item/clothing/cloak/half
-	if(prob(33))
+	else(prob(33))
```
JTGSZ, line 45: that needs to be if else(prob(33))

JTGSZ: the else despite the conditional being compiles, spits out a warning, and does its business as usual which means you will still not get the cloak = /obj/item/clothing/cloak/half/red as its being overwritten if the if(prob(33)) doesn't pass

Wintermote:
#755 was done to explicitly limit all the adventurer classes to 3 in their combat skills.

If we're going to go back on that we probably shouldn't be doing it piecemeal class by class, but rather rebalancing all of them appropriately.

GreatRuna:
> #755 was done to explicitly limit all the adventurer classes to 3 in their combat skills.
> 
> If we're going to go back on that we probably shouldn't be doing it piecemeal class by class whenever someone fixes something, but rather rebalancing all of them appropriately at the same time.

rogue has 4 in knives at current, I know that for a fact

YosemiteYam:
As Winter stated and as stated in your other PR bumping up adventurer skills, all adventurer skills were bought to 3 for a reason. If Rogues have 4 in knives its either an oversight or was done due to the fact that knives/daggers are objectively a worse main weapon than everything else. All this PR would do is make vaqueros experts at the same level as men at arms for absolutely no reason, "its their thing" isn't a reason.

PR #1011, opened 2024-05-23
Apple, butter and steak resprited.

Takyon69:
## About The Pull Request

Apple, butter and steak resprited.

image
image
image

## Why It's Good For The Game

idk


ophaq:
These look really good! I think the butter should be more softer/rounded looking, however like the current one.

Takyon69:
> These look really good! I think the butter should be more softer/rounded looking, however like the current one.

image

I tried that and didn't like the look of it.

PR #1012, opened 2024-05-23
Weaving & Silk

SELFHELL:
## About The Pull Request

This whole PR started with me being mildly annoyed that people could spam out half cloaks, allegedly expensive and exotic clothing used by nobility with almost no effort. I then asked myself why such a thing is made with fibers from bushes. The resulting realizations lead me to make this.

Adds:
Weaving Skill
Web Pants/Leggings
Web Shirt/Rags
webs
(slightly outdated, improved the shading/alpha on it)

Craftable Inquisitor/'Archivist" Shirt
The Silk Material

Rebalances:
Moves a bunch of church stuff to be under weaving and a couple of nobility items, didnt make sense for them to be made with fibers from a BUSH
Adds silk to the beespiders droplist

## Why It's Good For The Game

Makes clothing for the church/nobility more regal and labor intensive for the peasantry to create.
Also provides more usage for beespiders


cornercube:
minor note, the half-cloak is meant to be cloth fiber for poor people
the orphan spawns with it

regular black cloaks are on pilgrims, so that is also an average person outfit
the colored cloaks are the ones that might be silk, since those are the ones that spawn on people of slightly more importance

ophaq:
I don't like this much for the church stuff because it makes clerical robes too hard to craft. If there was an easier way of getting silk, maybe but I think it should be changed. It might be cool to have differing materials required to sew certain robes like Dendor hood/robes requiring fiber and wood/twigs, etc.

PR #1019, opened 2024-05-24
arrow proofing

cornercube:
## About The Pull Request
for some reason, maille provides literally zero protection to arrows as does gambeson and leather
and then plate armor has 100 bullet (arrows check on bullet resistance) and is immune

broadly increased protection from arrows, and by extension crossbow bolts, on non-plate armor by a small amount
most chain armor now has 20 bullet resist
leather armor has 10 bullet resist
gambeson is slightly stronger than maille at 25, for historical reasons
some armor that is described as plate but is not seemingly fully plated like chausses are 50
a few heavy pieces of clothing like leather gloves and leather vests get a trivial amount of arrow armor that will not do much in practice but can say they did something

bows dealing in excess of 50 damage with enough PER will now be dealing about half that against maille
crits are still possible through armor, so still not total protection, but that is just crit combat for you

## Why It's Good For The Game
why does a chunk of metal do nothing to stop arrows?
why does thick padded gambeson not stop arrows?
even if they can penetrate your armor, its going to at least slow down the arrow


ChungusGamer666:
will merge when conflicts are fixed

cornercube:
> will merge when conflicts are fixed

apparently there are concerns that overlapping armor might cause the weaker armor to be prioritized over the stronger one when you are hit, so we need to check that out

Gyrandola:
> > will merge when conflicts are fixed
> 
> apparently there are concerns that overlapping armor might cause the weaker armor to be prioritized over the stronger one when you are hit, so we need to check that out

I think I've encountered that bug

Real-MAGNUM:
Testmerging as per chungus' prerequisites

PR #1023, opened 2024-05-24
Making more things repairable, including containers.

Multigamer99:
## About The Pull Request

Added the ability to repair many more items than could originally be done:
- Added the ability to be repaired, via either hammer or needle, to all non-wooden items in modules/clothing/rogueclothes
  - Notably all cloaks, shoes, hats, masks, neckware, quivers, rings, shirts, belts, and bags.
 -  Added the ability to repair most non-wooden items, via either hammer or needle, within game/objects/items/rogueitems
 - Modified the behaviour of storage objects and needles in order to allow you to sew or hammer storage containers so you can repair them, with a bypass by clicking again on the storage container while you are sewing it if you wanna store your needle without repairing the container.

There are some flaws with this setup that I want to acknowledge:
- A quick double click doesn't store the needle: a short delay between clicks is required. I'll look into a fix for this if it's an issue.
- There is currently no way to store a hammer in a repairable container that is on the floor or on a table. It can be stored if you carry the container though, which should be fine enough as an option for players IMO, especially given the general lack of metallic containers.
- Some items can't be attacked, and thus can't be repaired. Most are not likely to be a concern for taking damage though and I had gone in and changed every item I've seen damaged while playing the game in this regard. That might have an unforeseen side effect though.
- I've only done limited testing on this so far, but it seems to work. I'm mainly sending this now so others can look at it and judge it since I have limited experience with dreammaker, and as a nice send-off point before I go to bed.

## Why It's Good For The Game

This was a change suggested in the discord that I decided to make for my first PR. It will make it actually possible to repair those pouches and cloaks you find that have taken damage, that way they look proper and new when examined and there isn't worry that you'll lose them to damage in the case of some things, like leather boots which used to be able to be destroyed and damaged in combat but could never be repaired. It's also not a huge change, so it shouldn't need too much reason beyond this: most of the items that were super important to repair before this were already repairable.

### DIFF
--- a/code/datums/components/storage/storage.dm
+++ b/code/datums/components/storage/storage.dm
@@ -553,7 +553,7 @@
 			var/obj/item/storage/this_item = parent
 			//Vrell - since hammering is instant, i gotta find another option than the double click thing that needle has for a bypass.
 			//Thankfully, IIRC, no hammerable containers can hold a hammer, so not an issue ATM. For that same reason, this here is largely semi future-proofing.
-			if(this_item.anvilrepair != null && this_item.max_integrity && !this_item.obj_broken && this_item.obj_integrity < this_item.max_integrity)
+			if(this_item.anvilrepair != null && this_item.max_integrity && !this_item.obj_broken && (this_item.obj_integrity < this_item.max_integrity) && isturf(this_item.loc))
 				return FALSE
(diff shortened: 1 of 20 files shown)


Multigamer99:
Reconfiguring my github because i'm a dumbass and set it up wrong. Will resubmit this later.

PR #1031, opened 2024-05-25
Makes most races have teenagers!!!

Multigamer99:
## About The Pull Request

What does this do?
- Makes most races that didn't have teenager as an option have it now.
- Affects Tieflings, Dark Elves, Elves, Dwarves, Aasimar, and Half Orcs (since they are still active in dev builds)
- Adds textures derived from their adult equivalents for Dark Elf men, Elf men, Dwarves in general, Half Orcs in general, and Aasimar/Tiefling Men.
- Adjusts item/body group offsets to fit with the new textures.

Potential concerns that may be present:
- All nude textures look fine, but I have yet to double-check my work on all skeletons.
- Most garments should fit fine from the test cases I ran but there is potential for an issue in there.
- Dwarf teens don't render beards properly at the moment, although the lack of beards may be considered a design choice rather than a bug.
- Some child roles may be good to have what races can fulfill them tweaked now that it somewhat matters.

Why am I submitting this now with these potential concerns?
- I have reached a spot I feel is suitably done.
- Any remaining bugs/errors I have found are ones I don't feel need addressing until someone who knows more about the lore weighs in.
- While I can bug test all I'd like ad infinitum, I won't catch every edge case nearly as fast as involving others can and it will likely need re-reviewed by others anyway.

## Why It's Good For The Game

For those in child roles, you are very little variety for character options. Additionally, it makes sense for the world to have youth of all the other races that aren't Humen or Half Elf since either the residents of Rockhill would have kids or youth might have emigrate to the town. It also allows newer players to play youth roles to try and have someone to teach them without forcing them to place one of two races, even if the other races tend to face extra challenges. More choice generally isn't a bad thing, and I think it's pretty fitting.


JTGSZ:
fsprites not entirely there but at least some more missing zeth design is filled in

PR #1034, opened 2024-05-25
Allows Roundstart Female Gaurds. No Sword Required.

InsightfulParasite:
## About The Pull Request
Allows Gaurds to be female.
In the bog we cant really deny any woman who is willing to become a meat shield for the town.
Im unsure if there is a reason that prevents female guards. If there is then ill take down the PR.

## Why It's Good For The Game
Women are allowed to be adventurers, bog guards, and paladins. Seemed weird that the patriarchy would just be in the royal court. Now veteran role players can complain in game about how they started letting women into the guard.

tweak: Watchman.


InsightfulParasite:
Fair enough.

PR #1041, opened 2024-05-26
QOL for Blacksmith crafting.

wgtjunior743:
## About The Pull Request

Added a new variable for the recipes called i_type

Added a new filter to anvil, when it verifies for the i_types on the game and if the ingot also has these types showing them if they exist, as the pictures, first one is gold and the second is iron.

Now before selecting the item you are going to make, you will first select it's type faciliting finding what you want to craft.

sc2
sc1
(It is upper case in the github code)

## Why It's Good For The Game

Before making any items we would have big list of abunch of mixed stuff, now it is has types and in the future it can have even more, new players will find where and what they want to craft easier, and the veterans will be able to select the item even faster.

### DIFF
--- a/code/modules/roguetown/roguejobs/blacksmith/anvil.dm
+++ b/code/modules/roguetown/roguejobs/blacksmith/anvil.dm
@@ -108,21 +108,37 @@
 	..()
 
 /obj/machinery/anvil/proc/choose_recipe(user)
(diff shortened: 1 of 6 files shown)


TheManWithNoHands:
Very nice. Will make it a lot easier for new people to pick up smithing.

PR #1046, opened 2024-05-26
Adds the crafting datum for Mantraps (Beartraps)

Joel-682:
## About The Pull Request: 
Adds the crafting datum for mantraps. A variant of beartraps that need time to set up, and can break under stress from doing so. The crafting recipe takes one small log (Subject to change) and uses the trapsmaking skill, with a difficulty of crafting of 1 (Subject to change, plus 10 int characters, plus 8 int characters with smokeweed and characters with novice trapmaking can make the trap). 

## Why It's Good For The Game

Adds another item into the game, with clear counterplay (Jumping over it/going around it/using the look around verb) and makes the trapmaking skill useful for making traps. Perfect for anyone who can't fight directly, as this will cover an escape (If you're quick enough to get it from your satchel) or at least slow down any would be pursuers.

### DIFF
--- a/code/game/objects/items/handcuffs.dm
+++ b/code/game/objects/items/handcuffs.dm
@@ -352,7 +352,7 @@
 				update_icon()
 				to_chat(user, "<span class='notice'>[src] is now [armed ? "armed" : "disarmed"]</span>")
 			else
-				user.visible_message("<span class='warning'>You couldn't get the shoddy [src.name] to open up!</span>")
+				user.visible_message("<span class='warning'>You couldn't get the shoddy [src.name] [armed ? "shut close!" : "to open up!"]</span>")
 /obj/item/restraints/legcuffs/beartrap/proc/close_trap()
 	armed = FALSE
 	alpha = 255
--- a/code/game/objects/items/handcuffs.dm
+++ b/code/game/objects/items/handcuffs.dm
@@ -352,9 +352,7 @@
 				update_icon()
 				to_chat(user, "<span class='notice'>[src] is now [armed ? "armed" : "disarmed"]</span>")
 			else
-				user.visible_message("<span class='warning'>The rusty [src.name] breaks under stress!</span>")
-				playsound(src.loc, 'sound/foley/breaksound.ogg', 100, TRUE, -1)
-				qdel(src)
+				user.visible_message("<span class='warning'>You couldn't get the shoddy [src.name] to open up!</span>")
 /obj/item/restraints/legcuffs/beartrap/proc/close_trap()
 	armed = FALSE
 	alpha = 255
--- a/code/modules/roguetown/roguecrafting/items.dm
+++ b/code/modules/roguetown/roguecrafting/items.dm
@@ -266,3 +266,11 @@
 	result = /obj/item/clothing/neck/roguetown/psicross/wood
 	reqs = list(/obj/item/natural/fibers = 2,
 				/obj/item/grown/log/tree/stick = 2)
+
+/datum/crafting_recipe/roguetown/mantrap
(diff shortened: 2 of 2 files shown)


Joel-682:
If the mantraps breaking after trying to be setup is too frustrating (It's a 50/50) it can be prevented by deleting the lines 354 through 357 and deleting the if altogether. Another solution may be to replace the deletion with a "You couldn't force the mantrap open!" message. For now it stays to punish players trying to use mantraps in the middle of combat, and it also limits mantraps to 5 per player rather than 5 plus how many they can scavenge/recycle mid-fight.

JTGSZ:
A lot of people have pointed out that you only want a single log to make a trap, and so will I.

Real-MAGNUM:
For every tree you can get something like 6 bear traps. Bad recipe. Make it require additional components like fiber, more logs, or anything else to lessen the huge spam potential.

Joel-682:
Very well, I'll go ahead and remove the code that breaks mantraps 50% of the time they're used, then. Doesn't feel fair to gather several resources and have them waste away.

Joel-682:
Made bear traps cost a rope (3 fibers, craft diff 1 IIRC) and two small logs. Can't think of any other requirements. I also went ahead and changed the code that broke the mantrap 50% of the time into one that simply doesn't let it open 50% of the time. If you're worried about spam, I'd like to remind you that attacking bear traps with any weapon disarms them.

Joel-682:
Made bear traps need iron, two fibers, and a single log, but now they make two beartraps per use. I can't test the changes yet, so don't merge this until I get it done/until someone else runs the tests.

Joel-682:
Yup, it's working. If these costs look good (1 Iron ingot, 2 fibers, 1 small log = 2 beartraps, requires table, craft difficulty requires trapmaking novice or plus 9 int) then it's ready to merge. Otherwise point out what you'd like changed.

code/game/objects/items/handcuffs.dm
```diff
@@ -352,9 +352,7 @@
 				update_icon()
 				to_chat(user, "<span class='notice'>[src] is now [armed ? "armed" : "disarmed"]</span>")
 			else
-				user.visible_message("<span class='warning'>The rusty [src.name] breaks under stress!</span>")
-				playsound(src.loc, 'sound/foley/breaksound.ogg', 100, TRUE, -1)
-				qdel(src)
+				user.visible_message("<span class='warning'>You couldn't get the shoddy [src.name] to open up!</span>")
```
JTGSZ, line 355: You went from a one stop end state to a singular message for a possibility of two different end states bro.
The prob is a failure, and it can occur whether it is armed or disarmed aka whether its open or closed.

Joel-682: Thanks for pointing that out, feeling like a dumbass. I'll fix it by making an if to change the message depending on if the trap is open or closed. Give me a second.

Joel-682:
Alright, fixed the text display, @JTGSZ, review it whenever you can, thanks again for the feedback.

CVSthePharmacy:
> Yup, it's working. If these costs look good (1 Iron ingot, 2 fibers, 1 small log = 2 beartraps, requires table, craft difficulty requires trapmaking novice or plus 9 int) then it's ready to merge. Otherwise point out what you'd like changed.

you've made it so complicated that now it's simply just not worth making

Joel-682:
It's a good getaway tool. And all you need is to buy some iron. The rest can literally be found by scavenging. Plus, it works on mobs. If you use this on a volf, spider, etc, they will fall down to one or two hits. Perfect with their predictable AIs. And again, they don't break now.

PR #1065, opened 2024-05-27
The Huddening- VampHUDS

Okram0:
## About The Pull Request
Vampires got HUDs now similar to the peasant rebels, I also moved a bunch of stuff from Chaosmode into its parent since a lot of stuff was runtiming and if I didn't nobody could become a vamp on any mode other than Chaosmode even with admin intervention. The runtimes would break the vampire lord on_gain shit.

## Why It's Good For The Game

Vampires SHOULD have HUDs showing them who the other vamps are, I've seen far too many cases of friendly fire, and it doesn't make sense that peasant communists can tell who each other are at a glance and vampires can't without an examine.


ChungusGamer666:
i dont like this at all, makes stuff too easy, and in hindsight it should be the other way around - peasant rebels should have to examine each other too

ChungusGamer666:
nvm vamps have telepathy so this only makes sense

PR #1074, opened 2024-05-28
[FIX] Repairable shields

Gyrandola:
## About The Pull Request

Shield banging had an oversight that made shields not repairable and.. only bangable.


CVSthePharmacy:
BANGABLE SHIELDS??????????

PR #1082, opened 2024-05-29
priest can no longer give out free spells and devotion holders

ChungusGamer666:
the server is moving in a low fantasy direction and spells should not be handed out on a frequent basis to random bums
priest can still convert people to acolytes and templars, but they do not get any of the spells, not even a devotion holder

### DIFF
--- a/code/modules/jobs/job_types/roguetown/church/priest.dm
+++ b/code/modules/jobs/job_types/roguetown/church/priest.dm
@@ -158,21 +158,6 @@
 	recruitment_message = "Serve the nine, %RECRUIT!"
 	accept_message = "FOR THE NINE!"
 	refuse_message = "I refuse."
-	charge_max = 200 //templars get cool spells, so higher cooldown
-
-/obj/effect/proc_holder/spell/self/convertrole/templar/convert(mob/living/carbon/human/recruit, mob/living/carbon/human/recruiter)
-	. = ..()
-	if(!.)
-		return
-	if(!recruit.cleric)
-		var/datum/devotion/cleric_holder/holder = new /datum/devotion/cleric_holder(recruit, recruit.patron)
-		holder.holder_mob = recruit
-		//Max devotion limit - Templars are stronger but cannot pray to gain more abilities beyond t1
-		holder.max_devotion = 250
-		holder.max_progression = CLERIC_REQ_1
-		holder.update_devotion(50, 0)
-		holder.grant_spells_templar(recruit)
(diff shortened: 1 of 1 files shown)


Xkallubar:
Low fantasy with wizards and vampires? I'd like to hear your definition of low fantasy.

ophaq:
Bruh you just ruined half the fun of the game. Now I can't do my many gimmicks like making a hospital or be a druid unless someone implements things to actually do said things (like an actual medical system, or druid class). Also the people have less incentive to join the priest now and church will be way more empty.

pilotmans:
Why.... did you do this? this is just another one of those prs that should be put on hold and not pr merged, wait until things get added that alleviate the damage this does, and if you're still unwilling, just make it so NOC cannot be used as an acolyte patron but instead automatically switches your patron upon join to either pestra or Astrata.

pilotmans:
I think i should clarify.
**This can Work**
But only if you add nuns and friars; at the same time, in order to add those you need to add things like chores to do inside of the church to justify their existence.

This is a band aid solution, and whilst it's more of a temporary solution compared to other options, everybody knows that temporary solutions are the most permanent of solutions

ophaq:
When are we unmerging this?

ChungusGamer666:
> When are we unmerging this?

Never. Giving out free spells is retarded. If you want spells, join as acolyte you dimwit.

ChungusGamer666:
> I think i should clarify. **This can Work** But only if you add nuns and friars; at the same time, in order to add those you need to add things like chores to do inside of the church to justify their existence.
> 
> This is a band aid solution, and whilst it's more of a temporary solution compared to other options, everybody knows that temporary solutions are the most permanent of solutions
This is literally all non sequiturs. What do friars and nuns have to do with not granting new converts free fucking spells?

ChungusGamer666:
> I get why you did this but until we actually have more fleshed out systems and roles I think this kind of thing should be put on hold. Also this server was already fantasy to begin with.
Most of the high fantasy elements were hastily shoved in by Raytown and not actually intended by Zeth in any manner. I do not agree with the D&Dslop direction they took and I think it was lazy, hence I will be course correcting.

ChungusGamer666:
> Low fantasy with wizards and vampires? I'd like to hear your definition of low fantasy.

Vampire: The Masquerade is not high fantasy.
https://tvtropes.org/pmwiki/pmwiki.php/Main/LowFantasy

Issue #1075, opened 2024-05-28
[REQUEST] It'd be nice if you could see whether or not people were still waiting to be revived inside of their bodies.


pilotmans:
Being buried or ignored due to people not realizing you've been waiting to be revived and aren't actually dead-dead, is quite annoying.

Solution: Upon examining a corpse that you've already checked the heartbeat of, you will see either "Their soul still hangs onto this plane" or "They are gone, they have already left this realm."

This would allow for a form of triage that rewards patient players, it would also be generally more player friendly cause it would mean being brought to the church isn't a 50/50 chance of whether or not you get revived or something stupid happens and you get left to rot.

ChungusGamer666:
This is very SS13 pilled but I could make this a passive trait for necrites and it would be nice...

PR #1083, opened 2024-05-29
New necrite passive: Soul examine

ChungusGamer666:
## About The Pull Request

Closes https://github.com/Blackstone-SS13/BLACKSTONE/issues/1075
Removes RTRAIT_ because it is fucking stupid, replaces all instandes with just TRAIT_

Necrites can now check heartbeat on corpses to know whether or not their soul has departed to the undermaiden's embrace.

## Why It's Good For The Game

1. It's more useful than not smelling decaying corpses
2. It's a fitting passive for necra worshippers
3. Not too hugboxy because only a few people, like morticians, will be doing it


ophaq:
This is a good change. Being immune to the stank would also be nice.

PR #1089, opened 2024-05-29
Riding skills

Real-MAGNUM:
## About The Pull Request

Gives riding skill to town guards, bog guard, bog master, village chief, and nobles.

## Why It's Good For The Game

With an eventual remap, bogtown will be separated from rockhill via anarchy zone. This will help the bog and town guard (castle guard already had skills) traverse between the gap at a bonus compared to most, if they buy saigas.

Indirect and intentional buff to butchers, since no one wants their stupid horses if they cant ride them well. Maybe bog guard will patrol outside the gate for once instead of letting bandits rogues and nudists camp the bridge? idk.


CVSthePharmacy:
give bog gaurds a stable next to the gate.... why else do they have saddles but no stables......

PR #1100, opened 2024-05-29
Makes a lot of things constructable, changes some recipes and crafting difficulty

PotatoTomahto:
## About The Pull Request
Adds a ton of constructible structures. Lots of verbage and description changes.
This is a balance PR because I messed with crafting difficulty to numbers that seemed reasonable to me. I also tried to make integrities of objects consistent and make sense.

**IMPORTANT:** Everything constructable should be breakable. Everything placed roundstart should not have changed functionality and be breakable. **TM THIS PR FIRST!**

## Why It's Good For The Game
More building good. Please yell at me if anything is broken or unbalanced.

### DIFF
--- a/code/_onclick/hud/screen_objects.dm
+++ b/code/_onclick/hud/screen_objects.dm
@@ -131,7 +131,7 @@
 			if(H.craftingthing && (H.mind?.lastrecipe != null))
 				last_craft = world.time
 				var/datum/component/personal_crafting/C = H.craftingthing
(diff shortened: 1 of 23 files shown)


AlamoTurtle:
PR not testmerged first, the no-longer-ubiquitous table is now impossible to create because the code deletes the entire recipe.

ophaq:
Now that this is going good, I would like to see more floor tiles, rugs, adjustable signs (bar/smith sign), bar counters, wooden lattice windows, etc. Great work despite some hiccups.

PR #1101, opened 2024-05-30
Diversifying Armour Into 3 Types

Warfan1815:
## About The Pull Request

This is a tenative PR, seeing as I'm not fully able to account for all the gameplay ramifications this will have. We'll probably need to first have some discussions, then a couple of TMs, and only then should we merge it.

There are now 3 diverse armour types:
-Blunt armour (better defined as area-pressure armour)
-Slash armour (better defined as line-presure armour)
-Stab armour (better defined as point-pressure armour)

For more information and discussion, see:
https://discord.com/channels/1209810478612873256/1245245799961591828


Wintermote:
I'm extremely wary (tentatively optimistic) of massive overhauls to combat mechanics; you've got years of iterative improvements that can easily be wiped out, leaving an high effort pet project that's less fun to play than what it's replacing.

In the discord discussion, my issues mostly mirror Potato's; the diversification of damage is potentially good, but the removal of armor penetration is not.

Warfan1815:
> I'm extremely wary (tentatively optimistic) of massive overhauls to combat mechanics; you've got years of iterative improvements that can easily be wiped out, leaving an high effort pet project that's less fun to play than what it's replacing.
> 
> In the discord discussion, my issues mostly mirror Potato's; the diversification of damage is potentially good, but the removal of armor penetration is (probably) not.

There certainly is a school of thought that prefers the tried to the untried. With how much opposition there is to deprecating armour penetration, I suppose I'll drop it in favour of simply adding the mechanics.

dunsel-hordon:
I think it's a smart move and actually forces different gameplay. Instead of one ore 2 weapons and intents being superior you would be able to mix and match your weapon and armor

ChungusGamer666:
Do not bring armor penetratoon back as it essentially invalidates most of this PR.

Warfan1815:
> Do not bring armor penetratoon back as it essentially invalidates most of this PR.

I'm going to need to rework crits before I can conclusively remove armour pen, it shall have to limp on yet.

ChungusGamer666:
Crits are handled entirely separate from damage calculations and use a variable called prevent_crits... Why exactly is that the case?

Warfan1815:
> Crits are handled entirely separate from damage calculations and use a variable called prevent_crits... Why exactly is that the case?

Crits should be tied to an increase in damage and probability, which leads then to what crits do at a moment. Crits can then subsume armour penetration, as a way to get through armour (with a chance.) If you think I should revert the revert of removing armour penetration I'll acquiesce.

ChungusGamer666:
this will cause every merge conflict under the sun so i'm carelessly speedmerging this

PR #1102, opened 2024-05-30
Dwarf Accent returns for a few days thanks to democracy vote of 85 to 23

JTGSZ:
Also for some reason there was a useless indent on a recently merged pr involving bundling worms, removes that.


ophaq:
We need to keep this forever. It's hilarious and no one knows what I'm saying half the time. It also saves me from having to type the account manually like I usually do anyway.

PR #1104, opened 2024-05-30
fixes zombie npc aggro

maaacha:
## About The Pull Request

makes zombies aggro by removing their factions. also adds them back when they lose zombie antag.

additionally adds a check to npc's equipping items to make sure they aren't zombies. because zombies are dumb and also can't hit things with a weapon.

## Why It's Good For The Game

zombie npcs work now


Gyrandola:
if this fixes zombie AI it's based.

PR #1123, opened 2024-05-31
Sconces have a craftdiff of 0, wall difficulty decreased

PotatoTomahto:
## About The Pull Request
Makes sconces and walls easier to craft. 

## Why It's Good For The Game
Here is your masonry trainer. Wall difficulty training was too high after doing some testing.


ophaq:
I miss floor tile training...

PR #1132, opened 2024-05-31
Makes dwarves have a seperate liver.

dunsel-hordon:
## About The Pull Request

I was trying to add drunken healing to the dwarf liver itself as apparently the quirk doesn't work. I could not figure out how to do it but i did make it so that dwarf liver is actually listed as such and is a mutant organ like elf eyes so that if someone wishes to add the drunken healing they may.
## Why It's Good For The Game
Lets someone skip a step when adding back drunken healing to the angry midget people.


TheManWithNoHands:
God yes. Slowly turning all racial traits and such into special organs you can harvest would be amazing.
Just imagine how the organ harvesting market will turn out.

PR #1136, opened 2024-05-31
prisonerGODS (they're noble hostages now)

E231kei500bandai:
## About The Pull Request

now that lunatic is the "punishment role" or whatever we dont need prisoners to be so shit and can make them actual worth playing and revert them to their original state of being hostages

## Why It's Good For The Game

makes way more sense realistically, random beggars nobody gave a shit about weren't kept as prisoners
more room for rp, the king can invite you to live in the keep or maybe negotiate your release


thebigtony:
I think there should be 2 slots instead of just one, but overall really good PR.

SpaceSlug2:
doesnt buff them to have 5 in wrestling like they used to, bad PR

PR #1140, opened 2024-06-01
Deletes the gobbo fort and puts it on its own map

dunsel-hordon:
## About The Pull Request
Fully deletes the gobbo fort and puts it on its own map file since people cheesed around the bedrock by digging though the ceiling.

## Why It's Good For The Game
No more free loot or food.


Real-MAGNUM:
Checks out, does exactly what I asked to be done.

PR #1141, opened 2024-06-01
the DEhuddening- Rebels

Okram0:
## About The Pull Request

Rebels no longer have a HUD telling them who other rebels are, though they can still examine to see their allies

## Why It's Good For The Game

pretty sure the maintainers wanted it, and so I might as well do it since it took me about 2 minutes to do. Might wanna TM first to see if rebel players will riot against coders

### DIFF
--- a/code/modules/antagonists/roguetown/villain/peasantrebel.dm
+++ b/code/modules/antagonists/roguetown/villain/peasantrebel.dm
@@ -51,14 +51,6 @@
 		if(new_owner.current && HAS_TRAIT(new_owner.current, TRAIT_MINDSHIELD))
 			return FALSE
 
-/datum/antagonist/prebel/apply_innate_effects(mob/living/mob_override)
-	var/mob/living/M = mob_override || owner.current
-	add_antag_hud(antag_hud_type, antag_hud_name, M)
-
-/datum/antagonist/prebel/remove_innate_effects(mob/living/mob_override)
-	var/mob/living/M = mob_override || owner.current
-	remove_antag_hud(antag_hud_type, M)
-
 /datum/antagonist/prebel/on_gain()
 	. = ..()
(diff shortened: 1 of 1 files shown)


Stagrawl:
Werewolves were literally just given HUDs.

Okram0:
Whats this got to do with werewolves? This is removing it from peasent rebels.

TheManWithNoHands:
Honestly would be a solution to the problem of:
"Runs in, yells recruit line, kills everyone that didn't accept."

wgtjunior743:
> Honestly, could be a decent solution to the problem of: "Runs in, yells recruit line, kills everyone that didn't accept."

facts maybe we will get some real reason to rebel besides, "JOIN!"

pismorizzo:
I would say Rebels should probably still get a hud to see who JUST the leaders are so people still know to listen to and protect them.

RicktenStone:
I feel like a better approach and more long lasting fix to rebel imbalance would be to also add a hud for loyalists, so when they reject join rebel speeches they get their own blue L icon that only they and loyalists can see. That way theres a point to reject rebellion as you will have a clear indicator of who is on your side and who isn't, since thats the biggest attraction of accepting rebel calls. 

This only just gets rid of the hud situation but doesn't make anything easier for loyalists/balance the gamemode.

YosemiteYam:
I feel like a better approach and more long lasting fix to rebel imbalance would be to also add a hud for loyalists, so when they reject join rebel speeches they get their own blue L icon that only they and loyalists can see. That way there's a point to reject rebellion as you will have a clear indicator of who is on your side and who isn't, since that's the biggest attraction of accepting rebel calls.

This only just gets rid of the hud situation but doesn't make anything easier for loyalists/balance the gamemode.

PR #1160, opened 2024-06-02
[FIX] No more deposits/withdrawals by "unknowns"

Gyrandola:
## About The Pull Request

Fixes being able to anonymously deposit or withdraw money from your bank account if your face is covered.

### DIFF
--- a/code/controllers/subsystem/rogue/treasury.dm
+++ b/code/controllers/subsystem/rogue/treasury.dm
@@ -160,10 +160,8 @@ SUBSYSTEM_DEF(treasury)
 ///@param character: The character making the deposit.
 ///@return TRUE if the money was successfully deposited, FALSE otherwise.
 /datum/controller/subsystem/treasury/proc/generate_money_account(amt, mob/living/carbon/human/character)
-	if(!amt)
-		return FALSE
-	if(!character)
-		return FALSE
+	if(!amt) return FALSE
+	if(!character) return FALSE
 	var/taxed_amount = 0
 	var/original_amt = amt
 	if(character in bank_accounts)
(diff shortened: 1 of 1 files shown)


code/controllers/subsystem/rogue/treasury.dm
```diff
@@ -160,10 +160,8 @@ SUBSYSTEM_DEF(treasury)
 ///@param character: The character making the deposit.
 ///@return TRUE if the money was successfully deposited, FALSE otherwise.
 /datum/controller/subsystem/treasury/proc/generate_money_account(amt, mob/living/carbon/human/character)
-	if(!amt)
-		return FALSE
-	if(!character)
-		return FALSE
+	if(!amt) return FALSE
+	if(!character) return FALSE
```
ChungusGamer666, line 164: hate let me tell you

code/controllers/subsystem/rogue/treasury.dm
```diff
@@ -177,26 +175,25 @@ SUBSYSTEM_DEF(treasury)
...
-        return
+/datum/controller/subsystem/treasury/proc/withdraw_money_account(amt, mob/living/carbon/human/character)
+    if(!amt) return FALSE
     var/found_account
     for(var/X in bank_accounts)
-        if(X == name)
+        if(X == character)
             if(bank_accounts[X] < amt)  // Check if the withdrawal amount exceeds the player's account balance
-                send_ooc_note("<b>The Bank:</b> Error: Insufficient funds in the player's account to complete the withdrawal.", name = name)
+                send_ooc_note("<b>The Bank:</b> Error: Insufficient funds in the player's account to complete the withdrawal.", character = character)
```
ChungusGamer666, line 189: i am almost certain that this is not right, i seriously doubt bank accounts are stored with direct references to a character instead of their name

ChungusGamer666 (changes requested):
oy vey

AlamoTurtle:
Doesn't the shylock ask for your finger to make an account? Would it not be using that instead of your face? Am I just missing some bug that allows you to make 70 alternative bank accounts?

wgtjunior743:
> Doesn't the shylock ask for your finger to make an account? Would it not be using that instead of your face? Am I just missing some bug that allows you to make 70 alternative bank accounts?

you can't make new accounts, but you can make one and later deposit / withdraw as unknown if you use a mask/hood

code/controllers/subsystem/rogue/treasury.dm
Gyrandola: That might have been changed when someone reported they could access some other guy's account with the same name. Anyways, I used the generate_money_account proc, aka the deposit one, as reference and it seems to work just fine.

ChungusGamer666 (changes requested):
Direct references to humans stored in the account list will create issues, I double checked the code. Do not mix and match strings with references to instances of human characters, all references should be strings - That is, the real name variable of the character.

PR #1161, opened 2024-06-02
Mostly an armor overhaul

MickDick1:
## About The Pull Request

First thing I did was increase mace 2h force to 30 from 20 (1h it's 20). Everything else is armor related I promise.

Conceptually:

HEAVY ARMOR PLATE: Have the best damage resists, but in the case that they are penetrated, offer no protection from wounds or lethal injuries.
HEAVY helmets got a durability boost of 2x (the ones that limited your FOV to just in front of you)

COMBINED HEAVY ARMOR: The best of both worlds, but much lower durability and slightly less resist than proper plate but critical resists rolled into one. Poorman's plate armor. It now requires gambesons to create once again.

MEDIUM ARMOR: Really only scale fits, and it was made to be like a really shitty plate armor (lamellar isn't that great) but it has incredibly economic efficiency. Unique compared to the chain-line of medium armor it has equal protection against ranged weapons as the rest of it's resists. Best paired with underarmor for inevitable damage penetration.
CHAINMAIL: Falling under the medium armor category, it has good durability and full slash resist, as well as cut/chop protections. Best paired with underarmor like before. STRENGTH: Chop resist

LIGHT ARMOR: sub-par resists. Anything with a bit of armor pen (daggers, spears, very high damage weapons) can generally get damage through this either partially or completely. However, it will prevent the worst effects of those weapons until it quickly falls apart. Best used by people who can eat pain or simply not get hit. Not terrible to wear with chainmail if you cannot acquire a gambeson to protect your vitals.

UNDERARMOR: Pretty much just gambesons/arming jackets and their head counterparts, coifs and arming caps (padded caps in sewing section). Full crit resists, but very little protection, very very low durability. Can save you in a pinch, but better worn under an outer layer. Can combined leather with gambesons now, as well as coif/padded cap can be worn with leather helmet. WEAKNESS: No chop resist
MASKS: they are taking the place of face-level underarmor but still being basically brigandine level with arrow weakness since there's no other way to make sure people don't just ignore everything else and aim eyes and knock you out in your great helmet.

Boots:  All boots that had significant armor now have twist resist. No more laying on the ground, breaking people's feet in half like a gremlin. 
Don't. Fall. Down. In. Combat.
Don't. Lose. Your. Weapon(s).

I didn't touch vampirelord forged armor stuff. So, it will still have its crit resists unlike normal plate armor but have more stab and blunt weakness. Seemed reasonable considering its ancient armor made with blood magic.

## Why It's Good For The Game

The purpose of this is to make combat less instantly lethal **if properly armored** and more about disabling someone so you can go for a coup de grace. This will probably make things worse before it gets better since run charge cycling exists and grabs can't be parried still but hopefully we will get there.

We really need an expanded inventory and UI to go with it. 2 layer for each part pretty much.
We really need normal wounds to generate when damage penetrates through armor that has that damage type's crit resist listed. (but still prevent critical ones)
Fix run charge funnies.
Fix grabs being unparriable.

This is almost certainly not the final draft. It probably won't even get merged. I'm just working with what we already have and how I understand it.
The chop resist thing was probably a mistake. But axes have shitty parry and must be 2h to actually start hurting you through plate normally, as well as a swing delay on chop that actually lets you pull it off. We will see.

If it does get approved, testmerge first.

### DIFF
--- a/code/__DEFINES/roguetown.dm
+++ b/code/__DEFINES/roguetown.dm
@@ -13,7 +13,7 @@
 #define PLATEHIT "plate"
 #define CHAINHIT "chain"
 #define SOFTHIT "soft"
-#define SOFTUNDERHIT "softunder"
+#define SOFTUNDERHIT "softunder" // This is just for the soft underarmors like gambesons and arming jackets so they can be worn with light armors that use the same sound
 
 /proc/get_armor_sound(blocksound, blade_dulling)
 	switch(blocksound)
--- a/code/__DEFINES/roguetown.dm
+++ b/code/__DEFINES/roguetown.dm
@@ -13,6 +13,7 @@
 #define PLATEHIT "plate"
 #define CHAINHIT "chain"
 #define SOFTHIT "soft"
+#define SOFTUNDERHIT "softunder"
 
 /proc/get_armor_sound(blocksound, blade_dulling)
 	switch(blocksound)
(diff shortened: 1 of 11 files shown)


CVSthePharmacy:
pr is massive steaming pile of shit, don't try to change the entire armor again

PR #1163, opened 2024-06-02
Decapitating people while the priest omen is active creates haunts

ChungusGamer666:
## About The Pull Request

The high priest is dead, millions must suffer

## Why It's Good For The Game

Think you're safe from the undead uprising because you decapitated your opponent, smartass? No you are not safe.


CVSthePharmacy:
is this instant or does it take a while and can I control the haunt

PR #1165, opened 2024-06-02
Update warrior.dm

BlackWombatty:
## About The Pull Request

Splits normal Warrior into Mighty Warrior and Agile Warrior.

Now with fixed equipment slots.

## Why It's Good For The Game

Warriors are in a weird place right now, They aren't actually better fighters than Rangers, Paladins, or the like - Even Bards have the same combat skills. This boosts them a bit, and lets them choose between heavy gear and strength, or light gear and speed. Either way, their combat skills are on-par with Templar and Men-at-Arms.

### DIFF
--- a/code/modules/jobs/job_types/roguetown/adventurer/types/combat/warrior.dm
+++ b/code/modules/jobs/job_types/roguetown/adventurer/types/combat/warrior.dm
@@ -21,33 +21,33 @@
 /datum/outfit/job/roguetown/adventurer/sfighter/pre_equip(mob/living/carbon/human/H)
 	..()
 	H.adjust_blindness(-3)
-	var/classes = list("Warrior","Monster Hunter",) // To Do - knight errant unique archetype(5 percent chance)
+	var/classes = list("Mighty Warrior","Swift Warrior",) // To Do - knight errant unique archetype(5 percent chance)
 	var/classchoice = input("Choose your archetypes", "Available archetypes") as anything in classes
 
 	switch(classchoice)
 	
-		if("Warrior")
+		if("Mighty Warrior")
 			H.set_blindness(0)
-			to_chat(H, "<span class='warning'>Warriors are well rounded fighters, experienced often in many theaters of warfare and battle they are capable of rising to any challenge that might greet them on the path.</span>")
-			H.mind.adjust_skillrank(/datum/skill/combat/crossbows, rand(1,2), TRUE)
+			to_chat(H, "<span class='warning'>Strong and tough, Mighty Warriors are ready to crush their opponents under their mace while wearing heavy armor.</span>")
+			H.mind.adjust_skillrank(/datum/skill/combat/crossbows, 2, TRUE)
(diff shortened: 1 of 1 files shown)


assortedbeads:
Shouldnt swift warriors get the new dodge trait?

BlackWombatty:
I've gone back and forth on that on my head. I don't know how much of a
difference it really makes, but they probably should. They get fewer stat
points than mighty warriors.

I'd like to hear other's opinions though!

Mortiz999:
I like the idea for making warriors more interesting, but the names don't really fit imo. I wrote up two alternatives if you want to use them:
Rename Swift warrior to Duelist: Duelists are lightly armored and evasive fighters, experts in executing precise strikes against man and beast.
Rename Mighty Warrior to Brawler: Strong, tough, and heavily armored, Brawlers overpower their opponents with might and metal.

CVSthePharmacy:
joebro... you reduced the swords skill on the heavier class... why? I thought you put them on par skill wise but no, you just made them apprentice in swords? like tf?????

YosemiteYam:
" Either way, their combat skills are on-par with Templar and Men-at-Arms."

- They're not supposed to be, adventurers are last place in the skill hierarchy of combat roles; Knights/Captain - Men at arms - Town Watch - Bog Guard/Adventurers - Non combat roles. Although the skills you gave them aren't actually making them on par with templar/men at arms so it should be fine.

- The agile/duelist version shouldn't start with studded armor. Studded is the best light armor you can get, it shouldn't be what you spawn with but work toward, give em normal leather or hide. 

- Strength bonus of 3 to the strong warrior is abit too insane, keep it at 2. They're already strong as hell and frequently end up stronger than the captain of the guard/knights with the right racial roll. 

- Might want to compare the swift "warrior" to the rogue traits in order to avoid just making a beefier rogue. Maybe not have them have the same speed, but the swift warrior has either slightly more str or endurance, since they are a warrior after all.

> joebro... you reduced the swords skill on the heavier class... why? I thought you put them on par skill wise but no, you just made them apprentice in swords? like tf?????

Probably cause the swifter class needs to be better at something in weapons than the other class otherwise its just a nerfed, lighter-armor version of a warrior, basically a weaker rogue. Maybe you'll understand how balancing works one day when your skull produces a brain to fill it.

BlackWombatty:
I agree with the armor, I'll change that.

As for the rouge thing - in a fight, they SHOULD feel like 'Rouge, but
better'. The things they trade for that are skills. They don't get sneak,
climb, pickpocket, and all those various things that let Rouges be the true
troublemakers they are.

My thought process is this - if a swift warrior and a rouge with equally
robust players just walk up to each other and start swinging, the warrior
should usually win. But if the flight starts with the rouge shooting the
warrior in the back of whatever their weakest-armored part is and then
closes into melee, the rouge should usually win. And the rouge has the
skills to pull that off more often than not, so long as they are even a
little careful.

PR #1143, opened 2024-06-01
More keys and drip. +assorted map fixes

TheManWithNoHands:
## About The Pull Request
Adds a new key for the royal bedrooms the queen now spawns with on her new keyring.
Court phys now also spawns with their key.

Gave the queen some more dress options allowing queens to dress up more roundstart. 
Same with court phys.

Assorted map fixes.

*You hear a Zad in the distance*

## Why It's Good For The Game
https://imgur.com/a/nndH4nz

The fact that the queen was unable to enter her own bedroom is stupid. So she now has her own keys to the bedroom, while the king can still enter it, being that his master key works for every lock. Don't set doors to its lock ID.

More drip all around.
Map fixes, and other little things.

PR #1170, opened 2024-06-02
Adds basic store/housefront with a loom, replacing the open table area

bowlofcereal:
## About The Pull Request

Someone asked me earlier to make a place with a loom so certain subjobs like the Seamstress could have a place to do business with the guildhall, but I thought it'd just be better to separate it somewhere else in town. Since literally nobody uses this spot for anything nor comes here, I guess now would be a good time to obliterate it and replace it with something more appealing. Just adds a building that starts with a loom inside of it with a spot for bartering next to it, because why not? Also tidies up the street next to it.

## Why It's Good For The Game

Nobody fucking comes to this spot of the map unless they are running away from someone or hiding BECAUSE no one comes here. We should try to make as many places in town as engaging as possible to incentivize people's interest.

Main
loom1

Second Floor
loom2

Roof
loom3


Stagrawl:
Beautiful stuff.

ophaq:
Nice, but it bugs me to no end that there is no tile in front of the top level stairway. You just exit off the side of the stairs. I know it would make the building one more tile wide but I think it would look better.

Real-MAGNUM:
> Nice, but it bugs me to no end that there is no tile in front of the top level stairway. You just exit off the side of the stairs. I know it would make the building one more tile wide but I think it would look better. Maybe try it for just the second floor so it's more like a tudor style with the top level being a bit wider?

Yeah, what he said. make it one tile wider to the left.

Real-MAGNUM:
Did it, and fixed some areas you messed up on accident.

Real-MAGNUM:
Merged your additions manually with #1143, closing this PR.

PR #1173, opened 2024-06-02
Adds cups/statues to matthios idol

MickDick1:
## About The Pull Request

Adds cups and statue items to matthios sacrifice list
made wooden mugs worthless

## Why It's Good For The Game

Bandits were unable to sell the valuable steel, silver, and gold stuff they found in these categories as they should.


MickDick1:
I have no clue what changing sellprice to 0 is doing to make it fail this hard. The details make no sense.

PR #1177, opened 2024-06-02
Fire damage multiplier reduced

PotatoTomahto:
## About The Pull Request
Reduces fire damage multiplier. Renames define.

## Why It's Good For The Game
Before: Pray smite kills you in about 5 seconds.
Now: Pray smite will reliably crit you but won't kill you.


Arkansas-Bar-Association:
pray smite should violently blast your guts out all over the pavement

PR #1180, opened 2024-06-03
Gives the Steward the Vault Key

efzapa:
## About The Pull Request

Steward's Keyring should now spawn with the proper vault key.

## Why It's Good For The Game

The main person responsible for all the money and treasures for the wealth horde having to bug the King for their king to place items inside was dumb.


wgtjunior743:
if you put it on the  feed the hole the item goes directly to the vault you don't need to open it, and the steward place upstairs has literally the vault key on the left closet and the right closet has other multiple keys, like bruh atleast look around before making the PR.

efzapa:
idk why they put the one key seperate from all the others, that is a bruh moment

PR #1181, opened 2024-06-03
Makes the Goldface and Nerve Master unlockable by the Master Key.

efzapa:
## About The Pull Request

Makes both the Nerve Master and the Goldface unlockable by the master key.

## Why It's Good For The Game

I feel like this should have been in in the first place, both for the sake of doing things if there the steward or merchant is gone/dead and they locked their machine, and to look at logs if the steward is being a shitter and stealing money.

### DIFF
--- a/code/modules/roguetown/roguemachine/merchant.dm
+++ b/code/modules/roguetown/roguemachine/merchant.dm
@@ -138,7 +138,7 @@
 /obj/structure/roguemachine/merchantvend/attackby(obj/item/P, mob/user, params)
 	if(istype(P, /obj/item/roguekey))
 		var/obj/item/roguekey/K = P
-		if(K.lockid == "merchant")
+		if(K.lockid == "merchant" || K.lockid == "lord")
 			locked = !locked
 			playsound(loc, 'sound/misc/beep.ogg', 100, FALSE, -1)
 			update_icon()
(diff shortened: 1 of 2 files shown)


MickDick1:
The nerve master of course, but not the goldface.
The merchant guild is a separate entity and would not accept the king having personal control of the business. If anything, the goldface needs to be unusable unless you're a certified member.

Gameplaywise we need to heavily guard against the king just killing the merchant and installing a new one/opening to public/etc.

efzapa:
> The nerve master of course, but not the goldface. The merchant guild is a separate entity and would not accept the king having personal control of the business. If anything, the goldface needs to be unusable unless you're a certified member.
> 
> Gameplaywise we need to heavily guard against the king just killing the merchant and installing a new one/opening to public/etc.

so what do you do when you want something and the merchant went out of their shop and ate shit, then died with their keys

MickDick1:
The guild would not set up without contingencies guarding against the hijacking of their economic conduit. Remember they have a whole balloon network and a teleportation device for the items that in the case of the local king disposing of their agents there would logically be disabled. Making it simple for the king to just unlock it just doesn't fit.

There is currently always their key you can just take off their corpse if they didn't hide it.

efzapa:
> The guild would not set up without contingencies guarding against the hijacking of their economic conduit. Remember they have a whole balloon network and a teleportation device for the items that in the case of the local king disposing of their agents there would logically be disabled. Making it simple for the king to just unlock it just doesn't fit.
> 
> There is currently always their key you can just take off their corpse if they didn't hide it.

presuming you can actually find said corpse

YosemiteYam:
Why would the goldface be unlockable. It's not even an item from the king's kingdom, it's a merchant guild importer so the king wouldn't be able to open it anyway since its from outside his kingdom.

efzapa:
@ChungusGamer666 ready now

PR #1184, opened 2024-06-03
Doubles heavy helm integrity and adds smash protection

MickDick1:
## About The Pull Request

Changes heavy helmet integrity to 400
Adds smash critical hit protection

## Why It's Good For The Game

These helmets cost 2 bars and completely fuck your FOV (you can only parry the guy right in front of you).
Everyone pretty much agrees they're worthless compared to just helmet + mask, which is collectively double the integrity without nearly the downsides.

This is also going to be the **only** helmet line that will protect you from being knocked out by shit like flails. Heavy helmets finally have an actual use case compared to alternatives.


CVSthePharmacy:
remove smash protection, also are you only buffing the pigface? really?

PR #1186, opened 2024-06-03
Necrite Shovel Knight

singularshamble:
## About The Pull Request
Gives Mortician an edge against zombies and skeletons when 2-hand-chopping with their shovel. Supposed to increase damage and create the same knockdown and stun effects that silver dagger does to werewolves and vampires.

Also makes the shovel part of the polearms group and gives Morticians Journeyman-level polearms skill to make them pretty good with their shovel.

Shovel is otherwise normal vs. other entities. Morticians that pickup another polearm are Journeyman-level with it, but only the shovel gets that gudgud.

## Why It's Good For The Game
Gives the Mortician:
1) A somewhat better defense against zombified corpses/ambush zombies and skeletons.
2) Another way to go about the round if they don't want to be just burying bodies and want to actively serve Necra by pre-emptively banishing undead before people are hurt
3) More of a reason to be brought along to the action by the Priest/the King when undead attack.

To wait around in/near the graveyard for the few hours that make up a round can sometimes be very void of activity; between burials Morticians often spam-pray (which I'm not saying is a problem), fend off people trying to steal dead-person loot, and waiting for the next body. By buffing the shovel a bit and only against undead mobs, the Mortician may now have the option to go out and either look for bodies or do more local burials in the Bog Town, and they won't be as easily killable by bodies they're trying to bury.

AT PRESENT:
The code works when I test it locally, but I don't see an improvement in damage vs the spawned in skeletons and zombies. I might be testing wrong. If someone could have a look and see if there's something I'm missing, I'd greatly appreciate it!!

Thanks!

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -224,3 +224,4 @@ tools/MapAtmosFixer/MapAtmosFixer/bin/*
 config/admins.txt
 cfg/admin.txt
 config/admins.txt
+code/rt.dm
--- a/code/datums/components/crafting/recipes.dm
+++ b/code/datums/components/crafting/recipes.dm
@@ -641,7 +641,7 @@
 	result = /obj/structure/headpike
 	category = CAT_NONE
 
-/datum/crafting_recipe/headpikebone
+/* /datum/crafting_recipe/headpikebone
 	name = "Spike Head (Bone Spear)"
 	time = 65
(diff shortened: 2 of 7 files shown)


CVSthePharmacy:
why did you close this

PR #1187, opened 2024-06-03
Athletic Skill Buff and XP gain

wgtjunior743:
## About The Pull Request

- Everytime someone run and lose stamina they gain abit of athletic xp based off their int, the max stam was based of only endurance  but now athletic also counts towards it, the skill level divided by 2 would be basically one single stat of endurance.

- Example : Level 6 skill of Athletic (Legendary) , Base Stat Endurance : X, on the max stamina calculation their Endurance will be XIII.

 
## Why It's Good For The Game

Athletic  only had calculations for knockback and was not trainable, now this skill can increase your stamina and is trainable, ofcourse it is slow but it is possible to reach lvl 3-4 with like 20-30 min training non stopping with rest pauses.
Technically a trainable stat but you really just increase the skill.

### DIFF
--- a/code/modules/jobs/job_types/roguetown/mercenaries/desertrider.dm
+++ b/code/modules/jobs/job_types/roguetown/mercenaries/desertrider.dm
@@ -64,6 +64,7 @@
 		H.mind.adjust_skillrank(/datum/skill/combat/whipsflails, 1, TRUE)
 		H.mind.adjust_skillrank(/datum/skill/combat/knives, 3, TRUE)
 		H.mind.adjust_skillrank(/datum/skill/misc/reading, 1, TRUE)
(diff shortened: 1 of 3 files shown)


GreenHoodiedEthan:
A good idea. I checked some of the jobs, and they don't start with athletics. Stuff like the mercenaries in particular don't start with the skill at all. It'd be a good idea to do a quick comb through and give some more roles the athletics skill. As it would make more sense for them to have.

wgtjunior743:
> A good idea. I checked some of the jobs, and they don't start with athletics. Stuff like the mercenaries in particular don't start with the skill at all. It'd be a good idea to do a quick comb through and give some more roles the athletics skill. As it would make more sense for them to have.

True , yeah i added, mercs have good training so ofcourse they have Athletics, because of a big skill change a while ago the adventures, and garrison got their skill fine.

PR #1198, opened 2024-06-03
miners n blacksmith FORTUNE PILLED???

Wirpy:
fortune-pills blacksmiths and miners since fortune affects mining and it's weird they don't get a buff to it. Hopefully it can incentivize actually going mining for once cause 9/10 times it appears that the Rockhill mine is empty, deserted and unused. (Im not calculing with dwarf fortresses here frfr thats diff man cmon) 

Makes mining a more viable job really and more likely fun as well since now you have a bigger chance to find gems and cool stuff!


ChungusGamer666:
good

PR #1202, opened 2024-06-04
Makes surgery tools craftable.

Multigamer99:
## About The Pull Request
Makes surgical tools actually craftable, albeit expensive for a full set, so that having your set stolen doesn't immediately remove your ability to do your job for the rest of the round.

Yes, this is a salt PR, but I do believe it is good for the game.

## Why It's Good For The Game
The new surgery mechanics aren't that useful because, well, noone can really use them. The two roles that have tools can easily get them stolen too. This gives them an option to replace their tools or have spares on hand in case, well, something happens and they lose theirs. However, a full set of tools is hard enough to get that they shouldn't become common after this change: It takes a LOT of steel to make a full set.

### DIFF
--- a/code/modules/roguetown/roguejobs/blacksmith/anvil_recipes/tools.dm
+++ b/code/modules/roguetown/roguejobs/blacksmith/anvil_recipes/tools.dm
@@ -125,6 +125,42 @@
 	created_item = /obj/item/rogueweapon/pick/steel
 	i_type = "Tools"
 
+/datum/anvil_recipe/tools/surgery/scalpel
+	name = "Scalpel"
+	req_bar = /obj/item/ingot/steel
+	created_item = /obj/item/rogueweapon/surgery/scalpel
+	i_type = "Tools"
+
+/datum/anvil_recipe/tools/surgery/saw
+	name = "Surgical Saw"
+	req_bar = /obj/item/ingot/steel
+	created_item = /obj/item/rogueweapon/surgery/saw
+	i_type = "Tools"
+
+/datum/anvil_recipe/tools/surgery/hemostat
+	name = "Forceps"
+	req_bar = /obj/item/ingot/steel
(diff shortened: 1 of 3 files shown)


Hmeister-real:
theres a reason they are rare, salt pr

Multigamer99:
Of course there is a reason they are rare, but completely irreplaceable and blocked off from the rest of the game save for two specific roles seems a bit excessive IMO. Surgery is quite limited in power anyway, compared to holy healing. I don't think this ruins the balance of the game and would probably help at least a few folks have more fun and pursue more RP possibilities than is possible without either this or something similar, such as them being a merchant purchase thing or a steward import.

Reyblood:
I don't think it'd be viable to make these ingame. Their costs are way too high (1 steel ingot per item).
You should add some descriptions to them.

ChungusGamer666:
Not a salt PR.

You forgot to change the smelt result of the surgery tools.

Multigamer99:
Ah, yeah, I can go back in and make them smeltable real quick once I get home from work.

As for the steep costs, the idea I had was that, clearly these are currently intended to be rare. As such, I didn't want to make them too cheap. That said, if you want them to be cheaper, I have a few ideas. Namely, Forceps making multiple at once, or even just a single recipe to make a whole kit (even if you can actually get by without some of the tools: Two forceps, A speculum, and Bone Forceps is enough to do surgery (forceps work as a scalpel and you can use other blades to cut bone iirc. Finally, incisions can be sewn and treated as any other wound.)) I was also thinking that a full set of tools for other things, such as farming, takes a decent bit of material too. Those three iron ingots are the same as four steel, so as it stands, a proper set of farming tools costs the same amount of raw metal ore as the bare minimum functional surgical tool set. I don't think that's too crazy balance-wise.

Multigamer99:
Turns out, I can edit code on my phone, so I made the change now.

Lemme know if I somehow fucked up changing a single line of text.

PR #1207, opened 2024-06-04
I'm retarded.

Stagrawl:
Makes druid a .dm

IDK, i'm retarded.


TheManWithNoHands:
Think I found out why that happened as well. The druid thing wasn't tickt.

CVSthePharmacy:
hi retarded, i'm dad

PR #1223, opened 2024-06-05
New bandit hideout location

MickDick1:
## About The Pull Request

Moves the bandit hideout entrance directly on the opposite end of the bog from the vampire one. It's not exactly subtle but neither was the original one with the fire right outside coming from the z level above.
Moved some adventurer spawns around so they aren't starting right next to it.
Closed up the original one, removing the old paths, etc.

## Why It's Good For The Game

The travel times from one end of the map to the other makes actually doing their thing too time expensive. You can make a couple trips and then the round ends. This should cut down a good chunk of it. All of the activity is pretty much just across the bridge, so being closer will cut down on travel times for the bandits who actually play the role.

Adventurers are all rushing to the gate and given the starting location of the bandits relative to all the adventurer spawns, they have no chance to actually rob anyone on the way to town before the entire playerbase is hiding behind the gate.

Test merge to see if I screwed something up and how it pans out preferably. Everything worked fine locally as far as I could tell.


Real-MAGNUM:
Looks great

PR #1228, opened 2024-06-05
50 prayer cycles per prayer click (100 devotion per)

Real-MAGNUM:
## About The Pull Request

Extends the amount of prayer cycles to be 50 cycles (100devotion) instead of 20 (40devotion)

## Why It's Good For The Game

Sit and wait simulator, not babysitting simulator.


CVSthePharmacy:
Can you make a little sound too so I know when its done?

PR #1232, opened 2024-06-05
I don't know what I'm doing but it works in testing

Stagrawl:
Allows Walter White to say his piece before the server reboots. 

Really you should just work this out yourself in a non-jank way.


Stagrawl:
@ChungusGamer666 save me, please, save me, please.

PR #1234, opened 2024-06-05
Halves leech drain

PotatoTomahto:
## About The Pull Request
Halves the base blood drain of leeches.

## Why It's Good For The Game
Leeches currently have double the original blood drain. When combined with the evil factor, it stacks into an absolute death sentence. This should help.


ChungusGamer666:
leeches range frlm 1 blood drain to 4 and fall out once they suck 122 units of blood
this pr feels quite hugboxy

PR #1236, opened 2024-06-05
weakens forge stuff for easier relocating

Real-MAGNUM:
title


ophaq:
Should do this to levers too.

PR #1243, opened 2024-06-06
Examining Armor or Clothing will now tell you what bodyparts it covers.

efzapa:
## About The Pull Request

Examining any body covering clothing or armor will now tell you what it covers/protects.

## Why It's Good For The Game

Sprites can sometimes be misleading, having a hard defined examine that tells you exactly what something covers can be good for people trying to figure out what armor may be best for them.


efzapa:
> Good idea, demonic implementation

will attempt to fix in a couple days

PR #1250, opened 2024-06-07
Fix sewage not making noise when you walk on it

### LINKED ISSUE #1008: [BUG] Sewage Not making noise when you walk on it
Issue reported from  Round ID: 404 (Blackstone (Rougetown on a Dedicated Server))

Reporting client version: 515.1637

Sewage water on the sewers makes no noise when you walk in, being able to be exploited to run away and make no sound

monad76:
## About The Pull Request
Fixes #1008 
Currently there is nothing that checks for when a player enters sewer water, this change simply adds that check and plays the appropriate sound for water movement. Since sewer water is shallow I omitted the splash sound effect on initial entry.

## Why It's Good For The Game
As stated in the issue players can run in water to completely mute themselves while being chased through the sewers which makes it much harder to track them given how dark it already is. Also everything should give some form of basic feedback to the player in some way.

### DIFF
--- a/code/game/turfs/open/water.dm
+++ b/code/game/turfs/open/water.dm
@@ -119,11 +119,10 @@
 			if(water_level == 2)
 				L.SoakMob(BELOW_CHEST)
 		if(water_overlay)
-			if(water_level > 1)
-				if(istype(oldLoc, type))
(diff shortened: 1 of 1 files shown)


BlueMemesauce:
If you're going to do this then just update it in the parent so that shallow water still gets a sound

monad76:
> If you're going to do this then just update it in the parent so that shallow water still gets a sound

yeah good point, it's been changed.

PR #1257, opened 2024-06-07
Minstrel - Towner Musician

Stagrawl:
Comes with instruments, has music skill for the future with either mood / bardistry.
 
Non-combat bard class.


Hmeister-real:
a little less athletics and reading, and up stealing by one

Stagrawl:
Eh, sure, I'm not married to their current skills. 

Their main thing is going to be relying on bardistry anyway eventually or RP.

Stagrawl:
At some point I need to fix the existing minstrel skills.

PR #1270, opened 2024-06-08
Church surgerybags

dunsel-hordon:
## About The Pull Request
These were previously added but maybe a mapping pr fucked it and took em out who knows but im adding them back. Also fixes a closed turf under merchant window.

## Why It's Good For The Game
Gives acolytes and church members actual tools to do surgery with seeing as they cost steel and map corretions.


ophaq:
Next we should give acolytes actual decent surgery skills so they don't constantly mess up and just resort to doing miracles because it's faster.

PR #1285, opened 2024-06-10
Enable some of the unabled and empty rare pilgrims

wgtjunior743:
## About The Pull Request

- New Legendary Mason
- New Legendary Carpenter
- New Legendary WoodCutter

- Nerf woodcutter carpenter from 5 to 1, no idea why that skill was so high, he cuts wood.
- Nerf carpenter carpenter from 5 to 4, like all other pilgrims it is 4 or less of his profession.

## Why It's Good For The Game

New pilgrims, people have another reason to pick pilgrim if they want to roll these jobs, more variety and possible fun roleplay.
I am also planning to enable some rare adventures but they will get their own gimmick.


WindowTalents:
I was just thinking about how much I want this to be something somebody does, nice!

I heard tiefling might be removed eventually, if that ever happens I want to make sure that Vaquero basically gets made into a role for another race (preferably not humen) so we can at least keep that role if we lose the silly demons.

I dunno the proper "lore" implications for that though.

Hmeister-real:
1 carp is far too low for a carpenter. make it pick (2, 2, 3) so theres a 2/3 chance for level 2

wgtjunior743:
> 1 carp is far too low for a carpenter. make it pick (2, 2, 3) so theres a 2/3 chance for level 2

Carpenter, carp level is 4 , the woodcutter carp level that is 1

PR #1292, opened 2024-06-10
Worms armageddon

ChungusGamer666:
wormy voices fixed i hate how byond pitch works


CVSthePharmacy:
Ok but can you add it so that when you die you explode too

PR #1306, opened 2024-06-11
Increases force when wielded from 20-30 for iron mace

MickDick1:
## About The Pull Request

force_wielded is changed from 20 to 30

## Why It's Good For The Game

Why does the steel mace get a damage increase from being 2 handed but not the iron variant? Not to mention all the other weapons.
It's just a clear inconsistency. Blunt is one of the worse damage types as well, given leather clothes/boots has quite high resists to it compared to others. No wonder everyone thinks this iron mace sucks ass. Because it does.

### DIFF
--- a/code/game/objects/items/rogueweapons/melee/blunt.dm
+++ b/code/game/objects/items/rogueweapons/melee/blunt.dm
@@ -1,6 +1,6 @@
 /obj/item/rogueweapon/mace
 	force = 20
-	force_wielded = 20
+	force_wielded = 30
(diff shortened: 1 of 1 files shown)


Wintermote:
The benefits of steel don't really apply to a mace; you just want something heavy and durable on a stick and they both do that.

This is a situation where someone made a weapon based on videogame logic instead of common sense. There shouldn't even be a distinction between iron and steel maces. 

This PR doesn't entirely resolve it, but it mostly resolves it, so I think it's good.

PR #1312, opened 2024-06-11
Goblin organs

Real-MAGNUM:
porting from

https://github.com/Rotwood-Vale/Rotwood-Vale/pull/40


Real-MAGNUM:
nvm organs in this context is related to bonercode that we dont have, it seems.

PR #1320, opened 2024-06-11
Zombies get no love from Xylix

DerFlammenwerfer:
## About The Pull Request
Recently, playable zombies were re-added, and it has resulted in a trend of people intentionally becoming zombies and subsequently using weapons and shields to go on low-risk murder sprees, whether because of boredom, because they simply want to be someone else's problem, or because they didn't roll the classes/antag they wanted.

Since players have been irresponsible with this recently returned toy and also to pave the way for re-implementing necromancer as an antag, I propose the following in order to discourage self-zombifying:

- When a player becomes a zombie, if their character's patron is anything other than Zizo, their luck is reduced to 0. This means that aside from bite attacks, or attacks on the weak and unprepared, they will not be inflicting critical hits, and will find it very hard to successfully use weapons. Attacking in larger numbers should be your go-to.
- If the zombified player IS a worshipper of Zizo, their luck is reduced to 5 instead. This is to allow self-zombifying by people who worship necromancy to act like a sort of challenge class. 
- There is a new trait now, called TRAIT_RAISEDEAD, which will eventually be applied by Necromancers when they resurrect a corpse. A zombie created from a corpse with this trait will have their luck set to 10, which is average.
- When a zombie becomes a rotman via Pestra's clerics, their luck is restored to the average of 10

## Why It's Good For The Game
Zombies lately have been the most common and most deadly antagonists in a given round, and becoming a zombie is as easy as drowning yourself in some random murk where you won't be discovered and decapitated or buried before you can rot. Combine this with the capability to use weapons or shields with infinite stamina means that players can self-zombify and get to play wish.com vampires. This harsh luck reduction is intended to curtail such behavior and force zombies to rely on their natural weapons or other strategy besides swift attack spam if they aren't raised by powerful necromancers.

### DIFF
--- a/code/__DEFINES/traits.dm
+++ b/code/__DEFINES/traits.dm
@@ -51,6 +51,7 @@
 #define TRAIT_IWASHAUNTED "iwashaunted" //prevents spawning a haunt from a decapitated body twice
 #define TRAIT_SCHIZO_AMBIENCE "schizo_ambience" //replaces all ambience with creepy shit
 #define TRAIT_SCREENSHAKE "screenshake" //screen will always be shaking, you cannot stop it
+#define TRAIT_RAISEDEAD "Necromancer's Binding"
 
 GLOBAL_LIST_INIT(roguetraits, list(
 	TRAIT_WEBWALK = "I can move freely between webs.",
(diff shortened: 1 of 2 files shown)


assortedbeads:
Zombies cant use weapons

emoats18:
If you don't mind to add me on discord I'd like to talk some mr coder 

emoats18

Real-MAGNUM:
Good PR.

pismorizzo:
Wanted to add it’s become an even bigger issue since Zombies now get a triumph upon infecting someone else. There’s mechanically incentive that if you are going to be killing people you might as well be a zombie so you get triumphs.

ChungusGamer666:
marked as DNM for now, it is likely my last zomble changes are enough

PR #1325, opened 2024-06-12
Vault strengthening

Real-MAGNUM:
## About The Pull Request

Buffs the vault's security by doubling integrity (2000) and changing integrity failure from 0.5 to 0.2, and triple thickening the outside walls, as well as walling off the broken gate. You MUST break at least one full masonry wall and a door, or two full masonry walls minimum. Three if you're trying to stealth it.

## Why It's Good For The Game

It's being rushed too much too quickly. Almost every round.

image


WindowTalents:
Absolutely blessed.

PR #1326, opened 2024-06-12
Restricts separated bogtown to 100 players, map rotation at roundend

PotatoTomahto:
## About The Pull Request
Readds Blackstone map with boggers next to gate as "Blackstone Dense". Map voting is called at the end of each round, and Blackstone with separated bog town requires 100 players minimum to be votable.

## Why It's Good For The Game
The new Blackstone map is designed only for highpop and feels miserable to play with less people.


TheManWithNoHands:
holup, let me wip you something up

wgtjunior743:
Truly feels like this, the seperation is Nice but with low pop ,everyone feels distant and interactions are harder to come by

CVSthePharmacy:
WAIT OLD BLACKSTONE MAP WITH THE BOG GATE GAURDING THE TRAVEL POINT???????????

CVSthePharmacy:
IF SO THEN WE'RE GOATED

PR #1330, opened 2024-06-12
Surgery bags

Real-MAGNUM:
Returns surgery bags to church


ophaq:
Now that we have town doctors we don't really need these. They barely get used anyway because miracles and the skill level required to reliable use them makes it difficult to use and you're better off just praying and doing miracles with the way things are currently.

PR #1335, opened 2024-06-12
Accomplished Physician trait for Court Physician and Barber Surgeon; Refillable needles

DerFlammenwerfer:
## About The Pull Request
The Accomplished Physician trait, **exclusive** to the Court Physician and Barber Surgeon, grants a percentage chance based on your medicine skill to not consume thread **while stitching wounds**. This chance is 20% starting at Apprentice level medicine, and increases by 20% per rank up to 100% chance at Legendary. This means that the Court Physician can effectively use any needle for his practice as though it were his Needle of Pestra.

Last but not least, per the suggestion of the Head Dev, the metal needles have 20 uses now, and thorns start with 5 but can have a max of 10. Refilling the needle takes 1 fiber per 5 uses, and requires you to touch the needle to the individual piece of fiber. The base time to refill it is 10 seconds, divided by your sewing skill + 1.

## Why It's Good For The Game
It has been my experience playing as a doctor that if you come across a patient who has suffered much blood loss, they may die before you can successfully perform surgery unless you first suture existing wounds. At times, it may also be preferable and faster to stitch an incision closed rather than damage your patient by cauterizing it; particularly if you just finished tending their wounds surgically. Also, with an influx of patients, you may not always have time to go ask a blacksmith for more needles or hunt down more thread without letting people die.

It also doesn't make much sense for needles to just be deleted when they run out of thread.


DerFlammenwerfer:
I don't know what you fucks did, but you thoroughly broke it and I have to start all over

PR #1344, opened 2024-06-13
Adds non-clerical version of Diagnose spell to Doctors/Court Physician

DerFlammenwerfer:
## About The Pull Request
This PR adds a non-miracle subtype of the Diagnose spell and grants it to doctors and the court physician. Since the Accomplished Physician trait will likely be removed in the future, and since two people commented on my last PR suggesting this be added, I've gone ahead and coded it.

## Why It's Good For The Game
It makes sense that dedicated doctor classes would know enough about anatomy to diagnose someone without needing the power of faith to do so, and it'll be a great QoL addition for the classes, since most doctors are already followers of Pestra anyway.


Stagrawl:
Spells are getting binned eventually so this is semi-pointless. Is there any way you can make this an interaction that relies on 3+ medicine?

WindowTalents:
> Spells are getting binned eventually so this is semi-pointless. Is there any way you can make this an interaction that relies on 3+ medicine?

Wait what? Yeah and just remove half the roles we have sure.

ophaq:
Finally, thank you for adding this. I have been screaming about this and playing these roles is ass without this.

PR #1345, opened 2024-06-13
Zombies can no longer run in the 90s

DerFlammenwerfer:
## About The Pull Request
Adds TRAIT_ZOMBIFIED, aka "Decayed Flesh," to zombies. This trait prevents a mob from activating run intent, and notifies them that they can't run anymore.

## Why It's Good For The Game
Nobody wants to have to deal with a hit and run zombie who sprints away at the speed of Plaid at the slightest resistance to just forever be a pain in the ass trying to fish for infectious bites all round.

Commit to the bit(e) or **super** die (again).


ChungusGamer666 (changes requested):
call it TRAIT_NORUN

DerFlammenwerfer:
> call it TRAIT_NORUN

Done and done.

PR #1351, opened 2024-06-14
Disables dungeon spawns

Real-MAGNUM:
AI is too buggy to make this anything other than a free loot generator.


WindowTalents:
I just wanna say, that beating the shit out of the goblins as a monk while they only sorta fought back (such as throwing them into each other and doing silly shit like pitting them) was quite a lot of fun when I was goofing around in there alone.

But, yeah, this will be super cool once the AI is threatening - it was great to see it.

PR #1353, opened 2024-06-14
Re-Adds Blacksteel Plate pieces for King

efzapa:
## About The Pull Request

I'm still not sure if this was intended but several people have told me this was mapslop and was not meant to be removed.

## Why It's Good For The Game

King actually having all their armor is pretty cool.


TheManWithNoHands:
I gave them the full armor set because I thought it looked cool. 
But it ended up being power-creep so i took away most of the blacksteel set.

Still, they end up with more than they had before. The chest armor + unique helmet.
Previously it was just the chest piece.

It looks cool, but its a complet full steel armor set for free.
Im bad at balance, but even i noticed the bad waves that caused.

PR #1372, opened 2024-06-15
Balances thievery around item weight class

DerFlammenwerfer:
## About The Pull Request
This is coded as an alternative to removing the ability to steal items in back slots, making the roll for stealing any item more difficult based on the w_class of that item; After beating the target's perception, to successfully steal the item(s) you want, you must beat (targetperception + (item w_class * 3). This means that if you do not have a certain level of thief skill, it is now entirely futile to try stealing items beyond a certain size, and even if you have the skill, there is a significantly higher chance of failing based on the target item's size.

## Why It's Good For The Game
Every rank of skill you have in stealing adds 1d6 to your roll vs. enemy perception. Since Perception is pretty much always less than 15, the law of averages means you can steal items without difficulty (such as gorgets, large weapons, backpacks or bevors) that you probably should have increased difficulty in stealing. This makes the system a more realistic, fair, and balanced.


Real-MAGNUM:
If possible, please check for previous bugs with pick-pocketing being impossible on sleeping targets due to cone of vision taking priority over closed eyes, or a sleeping state.

Real-MAGNUM:
Also this needs a balance check to make non-legendary pick-pocketing more viable, please.

DerFlammenwerfer:
> Also this needs a balance check to make non-legendary pick-pocketing more viable, please.

I have reviewed the law of averages for 6d6 and reduced the weight penalty accordingly, assuming 10 is the average perception. I also added a check for eyesclosed for FOV check purposes, excluded small and tiny items from the weight penalty, and threw in an auto-success for stealing from an unconscious target for good measure.

PR #1384, opened 2024-06-16
Disables Drifter queue

JTGSZ:
people are dming me about bugs, i do not have the energy to deal with bugs right now or dms about bugs


JTGSZ:
shit woops

PR #1388, opened 2024-06-16
BloatroleNERDS

bobjension:
fuck bogmen, fuck councillors and fuck bogmaster, physician is ok, inquistor is ok, baliff and captain of the guard needs to get merged back into sheriff.

### DIFF
--- a/code/modules/jobs/job_types/roguetown/courtier/councillor.dm
+++ b/code/modules/jobs/job_types/roguetown/courtier/councillor.dm
@@ -3,8 +3,8 @@
 	flag = COUNCILLOR
 	department_flag = NOBLEMEN
 	faction = "Station"
-	total_positions = 2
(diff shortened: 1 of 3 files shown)


WindowTalents:
The Bogmaster and the Bog Guards, are not bloat roles.

BUT THE COUNCILLOR IS - yes, I agree with ya there bud.

Coastfront:
In defense of councillor; It's nice to have a noble role thats relatively free to do gimmicks, compared to most court-roles having an abject responsibility.
The only issue is that the travelling-noble quite literally exists explicitly for this purpose, but just doesn't have manor access.

WindowTalents:
But, to be honest because of the lower pop - temporarily disabling these roles AND the bog town area entirely could be something worth looking at.

PR #1389, opened 2024-06-16
Merchant building changes

MickDick1:
## About The Pull Request

Replaces many wood walls with brick ones.
Moves goldface to the actual shop counters where either desk can easily access it.
Replaces merchant's room window with reinforced one
Adds second layer to merchant's gate at the balloon.

## Why It's Good For The Game

Brick walls are stronger and make the shop stand out more.
Goldface movement is a huge QoL benefit.
Merchant room having that weak window just forces you to build a new bed in a more secure section which is retarded.
Second gate layer at merchant's balloon is necessary for it to be properly secured against break ins.


TheManWithNoHands:
Just a small note. I get it with the walls. But like, you do realize the merchant building is the only building in town that uses dark wood walls. Rigth?

Also, i just now realised how fucked the wall choice in the keep is and i want to die.

MickDick1:
> Just a small note. I get it that you want the shop to stand out with other walls. But like, you do realize the merchant building is the only building in town that uses dark wood walls. Rigth?
> 
> Also, i just now realised how fucked the wall choice in the keep is and i want to die.

That wasn't the only reason. Wooden walls are the weakest walls in the game as well, brick has 25% more integrity as well as climb difficulty 4 vs wood 3, making it more secure.
You right though I didn't realize they were dark wood walls.

PR #1396, opened 2024-06-17
Alchemy slopfix try it

Hmeister-real:
## About The Pull Request

mayb works


Hmeister-real:
DONT MERG, NEED TO ADD MANAPOTS TOO. IL DO IT LATER

PR #1404, opened 2024-06-18
Enable drifters again, attempts to fix bugs.

PotatoTomahto:
## About The Pull Request
Report all bugs to me.


pismorizzo:
Perhaps Drifter should not be enabled seeing as it being enabled is why a significant amount of pop left Blackstone after how god awful it was, and was just a ploy to remove adventurers.

PR #1410, opened 2024-06-19
mild acolyte buff

MickDick1:
## About The Pull Request

Gave acolytes 2 in knives and polearms. 
Also 1 str, -1 pe, 1 spd.

## Why It's Good For The Game

What I often hear is acolytes are pretty much defenseless in spite of master wrestling and unarmed and the more I thought about it the more I sort of agreed. They're **retired clerics** who are done adventuring. Why do they have no weapon skills? Why are they so piss weak, given that context? So here is an attempt to rectify that a tad.


WindowTalents:
THEY HAVE WRESTLING FIVE WHAT DO YOU MEAN DEFENSELESS LOL

MickDick1:
> THEY HAVE WRESTLING FIVE WHAT DO YOU MEAN DEFENSELESS LOL

For starters, against people with armor you do 0 damage with punches/twists/smashes. Your only cope is to hold them down and slowly strip them which is not something that works outside of 1v1 combat. Wrestling has no parry roll, that's unarmed, and unarmed has no inherent defense stat like weapons do, so expert unarmed has the defense rolls of someone with apprentice weapon skill.

Not to mention monks are supposed to have fast reflex trait. But since they're retired, I didn't toss that in. Some very basic essential weapon skills make sense for clerics turned monks.

PR #1416, opened 2024-06-22
NEW MAP!?!!!

dunsel-hordon:
Makes a new map can use polish juices dry REVIEW IT.

## About The Pull Request
New map city size is 20 x 20 smaller along with drastically reduced bog town size and overall map size of 250x250. Also it is 6 layers.

## Why It's Good For The Game
Maybe it helps but map needs polish and decorations and input from others took altogether i think about 40 -50 hrs to make and 16hrs of that was done in one day. so ya.


dunsel-hordon:
Added cave system under the bog and a river through the map

Real-MAGNUM:
Looks sick. Still reviewing.

Real-MAGNUM:
Looks good enough for a testmerge to figure out all the bugs that certainly exist. You'll need to make a DM file to control the map I think, or whoever wants to actually enable it will have to.

PR #1418, opened 2024-06-24
Alchemy fixed

Ambrose997:
Fixes alchemy by ensuring the datums have unique names (they were being overwritten), corrects the crafting type at the top (was showing up in crafting instead of alchemy), more efficient usage of bottle subtypes amongst recipes via subtype_reqs = TRUE whilst also ensuring that it works regardless if you use obj/item/reagent_containers/glass/bottle or obj/item/reagent_containers/glass/bottle/rogue (this code uses both and the map has the non rogue version for some reason).

Stinky code freeze means players don't get to use this yet tho.

Tested this locally, fully works!

### DIFF
--- a/code/modules/roguetown/roguecrafting/alchemy.dm
+++ b/code/modules/roguetown/roguecrafting/alchemy.dm
@@ -1,8 +1,9 @@
-/datum/crafting_recipe/roguetown
+/datum/crafting_recipe/roguetown/alchemy
 	req_table = FALSE
 	tools = list(/obj/item/reagent_containers/glass/mortar, /obj/item/pestle)
 	verbage_simple = "mix"
 	skillcraft = /datum/skill/misc/alchemy
+	subtype_reqs = TRUE
 
 /datum/crafting_recipe/roguetown/alchemy/bbomb
 	name = "Bottle bomb"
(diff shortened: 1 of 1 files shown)


code/modules/roguetown/roguecrafting/alchemy.dm
```diff
@@ -1,96 +1,103 @@
...
+	craftdiff = 5 */
 
 /datum/crafting_recipe/roguetown/alchemy/health_pot_3x
 	name = "3x Health Potion"
 	result = list(/datum/supply_pack/rogue/food/healthpot)
 	reqs = list(/obj/item/reagent_containers/glass/bottle = 3, /obj/item/ash = 3, /obj/item/reagent_containers/food/snacks/grown/berries/rogue = 4, /obj/item/reagent_containers/food/snacks/fish/clownfish = 1)
 	craftdiff = 5
 
-/datum/crafting_recipe/roguetown/alchemy/health_pot_3x
+/* /datum/crafting_recipe/roguetown/alchemy/health_pot_3x_mana
```
Hmeister-real, line 87: mana_pot_3x

code/modules/roguetown/roguecrafting/alchemy.dm
```diff
@@ -1,96 +1,103 @@
...
+	craftdiff = 4 */ 
 
 /datum/crafting_recipe/roguetown/alchemy/health_pot
 	name = "Health Potion"
 	result = list(/obj/item/reagent_containers/glass/bottle/rogue/healthpot)
 	reqs = list(/obj/item/reagent_containers/glass/bottle = 1, /obj/item/ash = 1, /obj/item/reagent_containers/food/snacks/grown/berries/rogue = 1, /obj/item/reagent_containers/food/snacks/fish/clownfish = 1)
 	craftdiff = 5
 
-/datum/crafting_recipe/roguetown/alchemy/health_pot
+/* /datum/crafting_recipe/roguetown/alchemy/health_pot_mana
```
Hmeister-real, line 68: mana_pot

